PR #5, opened 2017-10-06
feat: Spinner component

wereHamster:
I saw that the spinner in the Sketch file I have for the discussion looks slightly different. But nothing should change on the component API. It should be a purely visual change.

image


tpreusse:
Looks good. One thing is missing though: exporting in `src/lib.js` for package users

wereHamster:
I missed that. It's there now.

PR #6, opened 2017-10-11
feat: Export raw font styles

wereHamster:
Export all raw font styles and document in the catalog.

Font style is a combination of font cut, size, line height etc. By collecting the list I already noticed that we use a few slightly-different sizes, eg. `sansSerifMedium22` and `sansSerifRegular21`.

In the Sketch file I have I also saw some use of font with size 15. I didn't add that style yet, because then we'd have sizes 14, 15, and 16 – three sizes which are very close together.

If I plot the font sizes which are currently used, then they follow quite closely the Major Third scale (each size is 1.25 times larger than the next smaller size). We might want to make that more explicit in the future.

image

### DIFF
--- a/src/components/Typography/docs.md
+++ b/src/components/Typography/docs.md
@@ -21,17 +21,70 @@ import {fontFamilies} from '@project-r/styleguide'
 </pre>
 ```
 
-A helper function to generate the `@font-face` css is available as `fontFaces`. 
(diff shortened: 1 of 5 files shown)


tpreusse:
Thanks and totally agree with not creating a 14, 15 and 16px cut.

PR #12, opened 2017-10-12
Comment components

wereHamster:
The components which render a whole tree will follow later. I have to reorganise the components I already wrote to support pagination.

But you can already use `<Comment />` for example on the profile page. 

image


src/components/Comment/CommentActions.js
```diff
@@ -0,0 +1,101 @@
+import React from 'react'
+import {css} from 'glamor'
+import {MdShare, MdKeyboardArrowDown, MdKeyboardArrowUp} from 'react-icons/lib/md'
+import {Label} from '../Typography'
+import MdMoreHoriz from '../Icons/MdMoreHoriz'
```
uxengine: FYI, this guy actually works for me:
https://github.com/gorangajic/react-icons/blob/master/md/more-horiz.js

src/components/Comment/docs.md
```diff
@@ -0,0 +1,83 @@
...
+```react|noSource,span-4
+<Comment
+  timeago='2h'
+  displayAuthor={{
+    profilePicture: '/static/profilePicture.png',
+    name: 'Paul Ullrich',
+    credential: {description: 'Bundesrat', verified: true}
+  }}
+  score={8}
+  content="This has to be the first time that an author who has written a Fantasy novel (The Buried Giant), is awarded with the Nobel Prize. He has also writted Sci-fi (Never Let Me Go), but Doris Lessing was first there (Shikasta). Wow, I did not expect this one. Don't get me wrong, I *love* his books, but he's way more mainstream than I expected."
```
uxengine: I'm not a lawyer, but just wanted to raise the question whether we should copy actual reddit comments or better use some Lorem Ipsum or self-created text?

src/components/Comment/docs.md
```diff
@@ -0,0 +1,83 @@
+### `<Comment />`
+
+A `<Comment />` is a combination of `<CommentHeader />` and the comment content. The actions are not part of this component.
+
+```react|noSource,span-4
+<Comment
+  timeago='2h'
+  displayAuthor={{
+    profilePicture: '/static/profilePicture.png',
+    name: 'Paul Ullrich',
```
uxengine: Maybe just use Christoph's real name instead of declaring a liquor dealer a Bundesrat? ;)

src/components/Comment/CommentActions.js
```diff
@@ -0,0 +1,101 @@
...
+    alignItems: 'center',
+    marginLeft: 10
+  })
+}
+
+export const CommentActions = ({score, onAnswer, onUpvote, onDownvote}) => (
+  <div {...commentActionsStyles.root}>
+    <button {...commentActionsStyles.replyButton} onClick={onAnswer}>
+      <Label>
+        Antworten
```
uxengine: Let's probably add a TODO for making hardcoded strings configurable.

src/components/Comment/CommentActions.js
wereHamster: Ok, I was using http://gorangajic.github.io/react-icons/md.html to see which icons are available.

src/components/Comment/CommentActions.js
wereHamster: Oh yeah, I wanted to pass this component `t` so that these strings can be localized form the outside.

src/components/Comment/CommentHeader.js
```diff
@@ -0,0 +1,66 @@
...
+    {profilePicture && <img
+      {...commentHeaderStyles.profilePicture}
+      src={profilePicture}
+    />}
+    <div {...commentHeaderStyles.meta}>
+      <div {...commentHeaderStyles.name}>
+        {name}
+        {timeago && <span {...commentHeaderStyles.timeago}>・{timeago}</span>}
+      </div>
+      {credential && <div {...commentHeaderStyles.description} {...(credential.verified ? commentHeaderStyles.verifiedDescription : {})}>
```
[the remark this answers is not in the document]
wereHamster: The reason why I did this is because the file contains two react components, and I wanted to make it clear which styles belong to which component. The `ActionButton` component seems to small to move into a separate file. But if the long variable name `commentHeaderStyles` is a concern then I can also inline `actionButtonStyles` into `commentActionsStyles` and rename the later.

src/components/Comment/CommentActions.js
wereHamster: Do you have a convention for translations keys? I'm thinking something that contains the path of the component would be useful, ie. `t('components/Comment/CommentActions/answer')`. What do you think?

If there is desire to keep the file hierarchy flat, then I could also move `CommentActions` one level up into `src/components/CommentActions`.

wereHamster:
@uxengine I think I addressed all the issues you have raised in the previous review.

PR #13, opened 2017-10-12
Add font styles needed for profile page

uxengine:
SerifBold{36,17}
screen shot 2017-10-12 at 15 37 01

SansSerifMedium{16}
screen shot 2017-10-12 at 10 30 17


uxengine:
Yes, switched to serifBold16 now.

PR #14, opened 2017-10-13
refactor: Upgrade react to 16

wereHamster:
One really nice feature of react v16 is that [components can return lists of elements](https://reactjs.org/blog/2017/09/26/react-v16.0.html#new-render-return-types-fragments-and-strings), without the need to wrap the elements in a container (eg. a div). I'd like to make use of that feature in the Discussion components.

Note that this is potentially a breaking change:

BREAKING CHANGE: Since react is a peer dependency, you may have to upgrade react in your own
project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   },
   "peerDependencies": {
     "glamor": "^2.20.37",
-    "react": "^15.4.2",
+    "react": "^16.0.0",
     "prop-types": "^15.5.7"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,13 +16,14 @@
...
     "cz-conventional-changelog": "^2.0.0",
     "glamor": "^2.20.37",
     "husky": "^0.14.3",
     "prop-types": "^15.5.7",
-    "react": "^15.4.2",
+    "react": "^16.0.0",
     "react-autocomplete": "^1.4.1",
-    "react-dom": "^15.4.2",
+    "react-dom": "^16.0.0",
     "react-maskedinput": "^3.3.4",
```
tpreusse: you can upgrade to [4.0.0](https://github.com/insin/react-maskedinput/releases/tag/v4.0.0). Not a React 16 release but no longer uses `React.PropTypes` and `React.createClass` – should be compatible.

PR #15, opened 2017-10-16
feat: CommentComposer component

wereHamster:
To show these components in the catalog I need a `t()` function. To make it convenient I created a a `docs.t.js` file in the component folder and expose that function to the whole catalog page using imports. But perhaps it would be better to store those strings in JSON files so that they can be used to seed the spreadsheet you have. And it would also allow us to show the required translation keys in the catalog pages.

image

### DIFF
--- a/src/components/CommentComposer/CommentComposerHeader.js
+++ b/src/components/CommentComposer/CommentComposerHeader.js
@@ -1,13 +1,15 @@
 import React from 'react'
 import {css} from 'glamor'
-import {MdCheck} from 'react-icons/lib/md'
+import {MdCheck, MdEdit} from 'react-icons/lib/md'
 import colors from '../../theme/colors'
 import {sansSerifMedium16, sansSerifRegular14} from '../Typography/styles'
 
 const styles = {
   root: css({
     display: 'flex',
-    alignItems: 'center'
+    alignItems: 'center',
+    background: colors.whiteSmoke,
+    padding: '12px'
   }),
   profilePicture: css({
     display: 'block',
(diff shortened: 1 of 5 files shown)


src/components/CommentComposer/CommentComposerHeader.js
```diff
@@ -0,0 +1,88 @@
...
+import {css} from 'glamor'
+import {MdCheck, MdEdit} from 'react-icons/lib/md'
+import colors from '../../theme/colors'
+import {sansSerifMedium16, sansSerifRegular14} from '../Typography/styles'
+
+const styles = {
+  root: css({
+    display: 'flex',
+    alignItems: 'center',
+    background: colors.whiteSmoke,
```
uxengine: Just for the record:
As discussed, these colors need to replaced by functional names once the color scheme is better defined:

src/components/CommentComposer/docs.t.js
```diff
@@ -0,0 +1,10 @@
+export default k => {
```
uxengine: Let's just follow the form component model and use component attributes like "placeholder" and "replyLabel", which will make the whole t() mocking obsolete.

wereHamster: This doesn't scale well. If you compose a few of these components into larger ones, you'll end up having to pass dozens of props to the top-level component, and it increases the likelihood of a name conflict. Already now we have two components which would require a `replyLabel`: `CommentActions` and `CommentComposer`.

uxengine: Good point about composition, I agree. So let's stick to your current version and defer further improvements of the translations setup to later.

PR #19, opened 2017-10-26
feat: use react-textarea-autosize to automatically expand CommentComposer

wereHamster:
Also auto-focus the textarea when the component is mounted.


tpreusse:
Let's add `react-textarea-autosize` as a peer-dependency. Will probably become a full one later – baked into `<Field />` and `<FieldSet />`

PR #20, opened 2017-10-31
Comment tree refactor

wereHamster:
I had to change the structure of the CommentTree components because the tree-like structure didn't work well with the sections that decrease the indentation when they are expanded. Now it's more list-like, where each comment is shown in a separate row. That works well thanks to the React 16 feature where render functions can return lists of element that don't need to be wrapped in an anonymous container (usually a div).

This is technically a breaking change (since the `CommentTreeComponent` API has changed), but nobody was using that component yet.

comment-tree-refactor


src/components/CommentTree/docs.md
```diff
@@ -1,61 +1,98 @@
-### `<CommentTreeLoadMore />`
+A discussion is logically a tree structure, but is displayed more like a list. This is so that we can display a sequence of direct replies without increasing indentation.
 
-```react|noSource
-<LoadMore t={t} count={128} onClick={() => {}} />
+Each comment is shown as a full-width div, with an appropriate number of vertical bars on its left to indicate its depth (`visualDepth`). Additional amount of left padding an be added with the `otherChild` flag. Below are three examples with varying visual depths, the last one has the `otherChild` flag set to true.
```
uxengine: Below are *two* examples ...

PR #23, opened 2017-11-01
feat: CommentTeaser

uxengine:
FYI, we'll need the title/subtitle version for rendering "latestComments" on the profile page. And while I'm at it, I also added support for the displayAuthor header (to be used when highlighting specific user comments in editorial context).

screen shot 2017-11-01 at 15 48 50

### DIFF
--- a/src/components/CommentTeaser/CommentTeaser.js
+++ b/src/components/CommentTeaser/CommentTeaser.js
@@ -1,0 +1,64 @@
+import React from 'react'
+import { css, merge } from 'glamor'
+import colors from '../../theme/colors'
+import { serifRegular14 } from '../Typography/styles'
+import CommentHeader from '../Comment/CommentHeader'
+import CommentTeaserHeader from './CommentTeaserHeader'
+import CommentTeaserFooter from './CommentTeaserFooter'
+
+const styles = {
+  root: {
+    marginBottom: '10px',
+    '& + &': {
+      borderTop: `1px solid ${colors.divider}`,
+      paddingTop: '20px'
+    }
+  },
+  body: {
+    ...serifRegular14,
+    margin: '10px 0'
+  },
+  clamp: {
+    // TODO: Replace with a crows-browser JS solution for line-clamping.
+    overflow: 'hidden',
+    display: '-webkit-box',
+    WebkitBoxOrient: 'vertical'
+  },
+  box: {
+    border: `1px solid ${colors.divider}`,
+    padding: '20px'
+  }
+}
+
+export const CommentTeaser = ({
+  t,
+  title,
+  subtitle,
+  displayAuthor,
+  content,
+  timeago,
+  commentUrl,
+  lineClamp,
+  isBox
+}) => (
+  <div {...css(merge(styles.root, isBox ? styles.box : {}))}>
+    {displayAuthor ? (
+      <CommentHeader {...displayAuthor} />
+    ) : (
+      <CommentTeaserHeader title={title} subtitle={subtitle} />
+    )}
+    <div
+      {...css(
+        merge(
+          styles.body,
+          lineClamp ? merge(styles.clamp, { WebkitLineClamp: lineClamp }) : {}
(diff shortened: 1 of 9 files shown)


src/components/CommentTeaser/CommentTeaser.js
```diff
@@ -0,0 +1,64 @@
...
+const styles = {
+  root: {
+    marginBottom: '10px',
+    '& + &': {
+      borderTop: `1px solid ${colors.divider}`,
+      paddingTop: '20px'
+    }
+  },
+  body: {
+    ...serifRegular14,
```
wereHamster: I made it a habit to always set color just below each font style, unless you want to explicitly depend on the color inherited from a parent element.

src/components/CommentTeaser/CommentTeaserHeader.js
```diff
@@ -0,0 +1,35 @@
...
+  title: css({
+    ...sansSerifMedium16,
+    color: colors.text,
+    whiteSpace: 'nowrap',
+    overflow: 'hidden',
+    textOverflow: 'ellipsis'
+  }),
+  subtitle: css({
+    ...sansSerifRegular14,
+    lineHeight: 1.1,
```
wereHamster: Why 1.1?

src/index.js
```diff
@@ -145,6 +145,18 @@ ReactDOM.render(
...
           {
+            path: '/components/comment-teaser',
+            title: 'CommentTeaser',
+            imports: {
+              t,
+              CommentTeaser: require('./components/CommentTeaser/CommentTeaser'),
+              CommentTeaserHeader: require('./components/CommentTeaser/CommentTeaserHeader'),
+              CommentTeaserFooter: require('./components/CommentTeaser/CommentTeaserFooter'),
+              CommentHeader: require('./components/Comment/CommentHeader')
+            },
```
wereHamster: In CommentTree I created a file `docs.imports.js` so that I don't have to edit this main index file each time I create a new component and want to document it in the catalog. That trick might be useful here, too, if more components are added.

src/components/CommentTeaser/CommentTeaser.js
```diff
@@ -0,0 +1,64 @@
...
+  title,
+  subtitle,
+  displayAuthor,
+  content,
+  timeago,
+  commentUrl,
+  lineClamp,
+  isBox
+}) => (
+  <div {...css(merge(styles.root, isBox ? styles.box : {}))}>
```
wereHamster: I noticed that you call `css()` here in the component render function instead of when constructing the `styles` object (eg. https://github.com/orbiting/styleguide/blob/e9ffc111659165b619e364d26e05050f5589ada6/src/components/Comment/CommentHeader.js#L7-L8). Is there any advantage / disadvantage?

src/components/CommentTeaser/CommentTeaserHeader.js
uxengine: Because of this:
screen shot 2017-11-01 at 17 24 24

The descenders (like the "g"s here) are cut off at a line-height of 1.0, because I use an "overflow: hidden" here for ellipsizing the potentially long subtitle. A line-height of 1.1 fixes that.

src/components/CommentTeaser/CommentTeaser.js
uxengine: Thanks for the tip, done.

src/components/CommentTeaser/CommentTeaser.js
uxengine: Actually, I've now removed css() completely here. "merge" just seems to be a semantic alias for "css" anyway, so it doesn't make sense to use both. The reason why I'm doing it in the render function is because I want to modify styles at render time depending on properties like "isBox".

src/index.js
uxengine: Cool, I've moved the imports to docs.imports.js now.

PR #24, opened 2017-11-01
feat: Overlay, OverlayToolbar

wereHamster:
localhost-3003-components-overlay

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -1,6 +1,6 @@
 import * as allMediaQueries from './theme/mediaQueries'
-export * as zIndex from './theme/zIndex'
 
+export {default as zIndex} from './theme/zIndex'
 export {default as colors} from './theme/colors'
 export const mediaQueries = allMediaQueries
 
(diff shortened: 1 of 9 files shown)


wereHamster:
I also added a module with zIndex values, so that we have a overview of all values that are used. Currently only one, for the overlay.

src/theme/zIndex.js
```diff
@@ -0,0 +1 @@
+export const overlay = 50
```
tpreusse: let's do a `export default {}` here like `colors` and `fontFamilies`. If we ever need to make this configurable it will be more painful with named exports

src/lib.js
```diff
@@ -1,4 +1,5 @@
 import * as allMediaQueries from './theme/mediaQueries'
+export * as zIndex from './theme/zIndex'
```
tpreusse: `export * as` is not part of the official module spec afaik – you need to import and export again

src/lib.js
```diff
@@ -19,6 +20,7 @@ export {Spinner, InlineSpinner} from './components/Spinner'
 export * from './components/Comment'
 export * from './components/CommentComposer'
 export * from './components/CommentTree'
+export * from './components/Overlay'
```
tpreusse: General thing: I prefer explicit exports here. This file should give an overview of the whole api surface.

Also in this specific case, as far as i understand, you need to name the default – `export {default as Overlay}`.

PR #26, opened 2017-11-03
Dropdown

wereHamster:
Using downshift was a pretty pleasant experience.

image

### DIFF
--- a/src/components/Form/Dropdown.js
+++ b/src/components/Form/Dropdown.js
@@ -1,0 +1,201 @@
+import React, {PureComponent} from 'react'
+import {css, merge} from 'glamor'
+import Downshift from 'downshift'
+
+import zIndex from '../../theme/zIndex'
+import colors from '../../theme/colors'
+import {sansSerifRegular21} from '../Typography/styles'
+import {Label, LButton} from './Label'
+
+const styles = {
+  root: css({
+    position: 'relative',
+    height: 40 + 20,
+    marginBottom: 12
+  }),
+  inner: css({
+    position: 'absolute',
+    top: 0,
+    left: 0,
+    right: 0,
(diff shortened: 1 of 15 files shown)


src/components/Form/Dropdown.js
```diff
@@ -0,0 +1,201 @@
...
+    position: 'relative',
+    height: 40 + 20,
+    marginBottom: 12
+  }),
+  inner: css({
+    position: 'absolute',
+    top: 0,
+    left: 0,
+    right: 0,
+    margin: '0 -12px 0 -12px',
```
uxengine: Optional:
Could be shortened to
```
margin: '0 -12px',
padding: '0 12px',
```

src/components/Form/Dropdown.js
```diff
@@ -0,0 +1,201 @@
...
+
+import zIndex from '../../theme/zIndex'
+import colors from '../../theme/colors'
+import {sansSerifRegular21} from '../Typography/styles'
+import {Label, LButton} from './Label'
+
+const styles = {
+  root: css({
+    position: 'relative',
+    height: 40 + 20,
```
uxengine: Directly use labelHeight and fieldHeight from Label.js and export there?

src/index.js
```diff
@@ -5,7 +5,7 @@ import {simulations, speedy, css} from 'glamor'
 import theme from './catalogTheme'
 import './global.css'
 import './catalogTheme.css'
-import {fontStyles} from './components/Typography'
+import * as fontStyles from './components/Typography/styles'
```
wereHamster: Why was this necessary?

tpreusse: can't set `speedy(false)` after `css` from `glamor` has been used.  `Typography/index` uses it

PR #30, opened 2017-11-09
Comment header overflow fix

wereHamster:
I noticed that we have a few places where we use the three css declaration to ellipsize text (white-space, overflow, text-overflow) so I extracted them into an 'ellipsize' mixin.

Descenders in the CommentComposerHeaders were cut off due to combination of ellipsize and line-height: 1. See screenshot below.

screen shot 2017-11-09 at 13 49 31

And I refactored CommentHeader and CommentComposerHeader to make sure the ellipsis only shortens the name/description and leaves the timeago and verified check always visible. There's now an example of an overflowing CommentHeader in the catalog.

image

### DIFF
--- a/src/components/CommentTeaser/CommentTeaserHeader.js
+++ b/src/components/CommentTeaser/CommentTeaserHeader.js
@@ -2,6 +2,7 @@ import React from 'react'
 import { css } from 'glamor'
 import colors from '../../theme/colors'
 import { sansSerifMedium16, sansSerifRegular14 } from '../Typography/styles'
+import { ellipsize } from '../../lib/styleMixins'
 
 const styles = {
   root: css({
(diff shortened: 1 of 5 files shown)


src/components/CommentTeaser/CommentTeaserHeader.js
```diff
@@ -11,17 +12,13 @@ const styles = {
...
     color: colors.text,
-    whiteSpace: 'nowrap',
-    overflow: 'hidden',
-    textOverflow: 'ellipsis'
+    ...ellipsize
   }),
   subtitle: css({
     ...sansSerifRegular14,
     color: colors.lightText,
     lineHeight: 1.1,
```
uxengine: That was my fix for descenders, we may wanna use your 20px lineHeight here as well for consistency.

PR #39, opened 2017-11-17
Editorial Base Template

tpreusse:
A Mdast Schema for Republik

Most important part: `src/templates/Editorial/index.js`

See documentation live: https://r-styleguide-pr-39.herokuapp.com/templates/editorial
screen shot 2017-11-23 at 19 17 27


src/templates/Editorial/index.js
```diff
@@ -0,0 +1,392 @@
...
+  InfoBoxFigure
+} from '../../components/InfoBox'
+
+import {
+  matchType,
+  matchZone,
+  matchHeading,
+  matchParagraph,
+  matchImageParagraph
+} from 'mdast-react-render/lib/utils'
```
lukasbuenger: Is there a reason why `mdast-react-render` is not listed as dependency?

src/components/PullQuote/Text.js
```diff
@@ -12,13 +12,21 @@ const styles = {
...
+    }
   })
 }
 
 export const Text = ({ children, attributes, isQuoted = true }) => {
   return (
-    <div {...attributes} {...styles.text}>
-      {isQuoted ? <span>«{children}»</span> : <span>{children}</span>}
+    <div {...attributes} {...styles.text} {...((isQuoted && styles.quoted) || {})}>
+      <span>{children}</span>
```
uxengine: The `<span>` is obsolete now.

src/templates/Editorial/index.js
```diff
@@ -0,0 +1,392 @@
...
+                  component: PullQuoteSource,
+                  editorModule: 'paragraph',
+                  editorOptions: {
+                    type: 'QUOTECITE',
+                    placeholder: 'Quellenangabe / Autor'
+                  },
+                  rules: [
+                    br,
+                    link
+                  ]
```
uxengine: No action required now, but just for the record:
As a left-wing, this code is veering way too much to the right IMHO ;) 
We might wanna consider breaking things down into smaller objects, and composing the main schema from these.

lukasbuenger: Word

PR #41, opened 2017-11-20
fix: Autocomplete based on VirtualDropdown and downshift. Resolves #40

### LINKED ISSUE #40: AutosuggestField not usable in its current form
As of now, this component delegates to react-autocomplete while imperatively dumbing its feature catalogue down to the most simplistic of use cases, which you hardly ever come across when dealing with autosuggestion problems in real life. Even plain HTML selects are more flexible.

I'd expect a styled proxy for react-autocomplete, instead I get a fancy string-only select. In its current form not a worthy addition to this library. 

I will try to solve this in an upcoming PR but, tbh, was quite astonished by the fact that this got  released in the first place.

lukasbuenger:
Based on #40 

#### VirtualDropdown
* Exports `styles`
* Exports `itemToString`
* `<ItemsContainer />` updates its height after each update
* `<Items />` passes the index to `getItemProps`
* `<Items />` performs value equality to determine if selected.

#### AutosuggestField
* Removed

#### Autocomplete
* New element

#### Docs
* Section about AutosuggestField removed
* Section about Autocomplete added

#### index.js
* Replaced AutosuggestField with Autocomplete inclusion

#### lib.js
* Replaced AutosuggestField with Autocomplete export

#### TODO
* Maybe we should use `itemToString` everywhere. This would help with arbitrary item types.
* Find a reasonable way to tell `<ItemsContainer />` to update its height only if necessary.
* Talk about async helpers, e.g. `<Loaders />` and such. 

image
image
image


src/components/Form/Autocomplete.js
```diff
@@ -0,0 +1,82 @@
...
+        return <div {...styles.root}>
+          <Inner isOpen={isOpen}>
+            <Field
+              label={label}
+              renderInput={
+                ({ onBlur: onFieldBlur, ...fieldProps}) => {
+                  return <input {...{
+                    ...fieldProps,
+                    ...getInputProps({
+                      onBlur: (...args) => !!selectedItem || onFieldBlur(...args),
```
tpreusse: let's make if explicit – `if (!selectedItem) { onFieldBlur(...args) }`

tpreusse: but why only forward it when not selected?
screen shot 2017-11-22 at 13 38 37
there is currently a bug the first time one select something—probably a race condition where the blur goes through.

I would aim for a solution that always forwards the blur and the value (label of selectedItem) – not only the filter

Also currently one can't unselect, deleting everything in the field and blurring should trigger a `onChange` with undefined / null.

lukasbuenger: I really can't reproduce the bug. Could you elaborate (browser, exact behavior)? 
Will respond to the suggestions a little later

tpreusse: @lukasbuenger chrome 62, open https://r-styleguide-pr-41.herokuapp.com/forms#autocomplete, click field, type `J`, select `Januar`

Only happens the first time when there isn't a selected item

tpreusse: also happens in safari 11 and ff 52

lukasbuenger: Got it. Almost never use the mouse for that kind of selection. Thanks!

lukasbuenger:
I was able to fix the focus AND get rid of the manual checking and piping by preventing default onMouseDown on the Dropdown internal `<Item />`. This, because the label position is based on css and downshift propagates its change only after a complete click cycle.  Please check again.

lukasbuenger:
A for this one:

> Also currently one can't unselect, deleting everything in the field and blurring should trigger a onChange with undefined / null.

Two reasons why I really think this is a bad idea. 
1. With no further addition, the Autocomplete behaves exactly like 
```
<select id="foo" name="select">
  <option disabled selected>Choose</option>
  <option value="value1">Value 1</option>
  <option value="value2">Value 2</option>
  <option value="value3">Value 3</option>
</select>
```
There is no way of not having a value, once you initially chose one. Without the disabled option it would assume the first item as initial, which would not be very clever considering that we're most likely going to use this to filter server responses, which may or may not be available when initializing the widget. I think in its purest form, this is how a bare bone autocomplete should work: Its a select with searchable options. This is established UX.

2. If I had to implement the current default behavior based on a widget that acts the way you suggested, I'd have to prevent it from losing state instead of declaratively giving it properties. While the other way round, I'd declaratively would tell it, that it has none when I consider my `filter` value to represent an empty selection. 
In my opinion you suggest implicit internalized behavior that is quite quirky to circumvent which leads to bad dev experience.

src/components/Form/Autocomplete.js
```diff
@@ -0,0 +1,76 @@
...
+        highlightedIndex,
+        isOpen
+      }) => {
+        return <div {...styles.root}>
+          <Inner isOpen={isOpen}>
+            <Field
+              label={label}
+              renderInput={
+                fieldProps => (
+                  <input {...{...fieldProps, ...getInputProps()}} />
```
tpreusse: currently focus and blur are not propagated to `<Field />` which leads to incorrect label states.

Could you forward them by doing `{...getInputProps(fieldProps)}`—see [downshift readme](https://github.com/paypal/downshift#getinputprops)—or is there a reason why not?

Incorrect states after blur:
screen shot 2017-11-22 at 20 15 21
Steps: Click field, click else where
Expected: Label animated back to placeholder look

screen shot 2017-11-22 at 20 16 34
Steps: Click field, select Januar, click else where
Expected: Label turns grey

lukasbuenger:
> Render the selected value label in grey on top of the field when there is one, isOpen and filter is empty

I used a placeholder on the input. The changing label was rather confusing. What do you think?

> Add a clear X by default (import MdClose from 'react-icons/lib/md/close')

I can do that. But what about we offered some kind of "icon slot" in the props? Then you could add any icon or "after" elements you'd want. Because, again, I'd rather don't have this ship with too much out of the box internally.

tpreusse:
Let's pair on this tomorrow to figure out the focus issues (can't use `simulate`, powered by a glamor dev helper) and get it merged. Placeholder sounds good. And a icon slot on `<Field />` is probably a missing feature anyhow.

PR #45, opened 2017-11-27
Link, Emphasis, Refactor & Clean Up

tpreusse:
Changes
- mv `Editorial.AuthorLink` to `Editorial.A`
- add `Editorial.Emphasis` aka bold
- rm regular breakout size (just use undefined)
- unify figure exports

And some clean up and doc fixes.

### DIFF
--- a/src/components/Center/docs.md
+++ b/src/components/Center/docs.md
@@ -1,7 +1,7 @@
 `<Center />` defines a max-width and ensures a horizontal padding.
 
 `<Breakout />` allows to adjust `size` (and positioning) of its content:
-- `regular` (default)
+- `undefined` (default, do nothing)
 - `narrow`
 - `tiny`
 - `breakout`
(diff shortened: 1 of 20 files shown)


tpreusse:
@lukasbuenger can you be more specific what you're missing? All planned variances of `<Figure />` are documented with an example on the [figure page](https://r-styleguide-pr-45.herokuapp.com/figure).

lukasbuenger:
After
> A `<Figure />` typically contains an `<FigureImage>` and an optional `<FigureCaption>`. The `<FigureImage>` component makes sure that the image doesn't exceed the container width when using a large image.

I'd expect something like

Properties:

**`<FigureImage />`**
- `data` - A dictionary containing keys
    - `src` - String, the image url, mandatory
    - `alt` - String, the alternative text

tpreusse:
@lukasbuenger added it. But let's also keep PRs focused.

I think generally documentation coverage in the styleguide is pretty good and we have to balance our energy between examples, prop types and documentation lists of the same.

lukasbuenger:
@tpreusse 
I really think, that it doesn't take too much time to spec and document a prop type. If that requirement does indeed make your PR's less focused, well ...

PR #56, opened 2017-12-08
fix(Figure.Image): Reserve expected space while loading and support image resize

uxengine:
For image URLs with sizing info.

While loading, the image itself is used as a placeholder with gray background and paddingBottom/height:0 sizing. When the image has loaded, we move back to the default image size at 100% width.

While image is loading:
screen shot 2017-12-08 at 16 55 34

Image has loaded:
screen shot 2017-12-08 at 16 55 40

### DIFF
--- a/src/components/Figure/Image.js
+++ b/src/components/Figure/Image.js
@@ -1,84 +1,51 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
-import { css, merge } from 'glamor'
-import { parse, format } from 'url'
+import { css } from 'glamor'
+import { imageSizeInfo, imageResizeUrl } from 'mdast-react-render/lib/utils'
 
 const styles = {
   image: css({
     width: '100%'
   }),
-  imageLoading: css({
+  wrappedImage: css({
+    position: 'absolute',
+    width: '100%'
+  }),
+  aspectRatio: css({
     backgroundColor: '#eee',
-    height: 0
+    display: 'block',
+    position: 'relative'
   })
 }
 
-const imageSizeInfo = url => {
-  const urlObject = parse(url, true)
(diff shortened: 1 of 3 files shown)


src/components/Figure/Image.js
```diff
@@ -1,20 +1,91 @@
...
+        {...attributes}
+        {...css(
+          styles.image,
+          aspectRatio && loadStatus === 'loading'
+            ? merge(styles.imageLoading, {
+                paddingBottom: `${100 / aspectRatio}%`
+              })
+            : {}
+        )}
+        onLoad={this.handleLoaded}
```
tpreusse: `onLoad` is not supper reliable and behaves differently depending in the browser engine, e.g. only the first time a image loads:
https://bugs.chromium.org/p/chromium/issues/detail?id=7731#c12

I'd prefer to go with a stateless solution:
```
<span {...styles.aspectRadio} style={{paddingBottom: `...`}}>
  <img />
</span>
```

A `span` with `display:block;` images are inline blocks and can be used where `div` aren't strictly allowed.

Alternative would be to handle changing `src` props, add `onError` handle and put a `key={src}` on the image to tell react to not re-use the dom node when src changes.

tpreusse: I've been thinking about this some more: our images will become stateful for other reasons—proper lazy loading on scroll, srcset fallback for old browser or a polyfill implementation—don't spend any time making it stateless just for the sack of it. The `onLoad` concern are still there though.

uxengine:
Comments addressed as discussed.

Works well on FrontImage and FrontSplit teasers, with and without size info.
Will need some extra work for FrontTile teaser, since it currently uses a generic  element. Quick check showed it's not completely trivial to move things over to Figure/Image

PR #57, opened 2017-12-11
Port Front Schema

tpreusse:
- add link support
- safety for figure image src
- rename `TeaserFrontAuthorLink` to `TeaserFrontCreditLink`—sorry I couldn't resist
- fix credit link props passing (was not passing href/title)

Includes basic docs, could be extended by the schema author ;)
screen shot 2017-12-12 at 00 54 53


src/templates/Front/index.js
```diff
@@ -0,0 +1,389 @@
...
+    editorModule: 'headline',
+    editorOptions: {
+      type: 'FRONTFORMAT',
+      placeholder: 'Format',
+      depth: 6,
+      optional: true
+    }
+  }
+
+  const frontImageTeaser = {
```
uxengine: Thanks for extracting these rules out of the otherwise huge schema object.

It might help readability to name these
frontImageTeaser*Rule*
frontSplitTeaser*Rule*
etc. to make clear at declaration time how they're gonna be used further down.

PR #59, opened 2017-12-14
feat: VideoPlayer

uxengine:


uxengine:
Mostly copied as is from Crowdfunding/republik-frontend, with minor renaming changes ("hidePlay" is now "showPlay", "poster" is now "thumbnail").

Preview: https://r-styleguide-pr-59.herokuapp.com/videoplayer

PR #66, opened 2017-12-22
Figuregroup

lukasbuenger:
Figurecaption module instead of paragraph module


tpreusse:
@lukasbuenger the squash merge you performed causes it not to release because you've used a message without the correct [commit message format](https://github.com/orbiting/styleguide#semantic-release).

PR #71, opened 2017-12-30
fix(Article): Don't set maxWidth on an edge-to-edge cover image.

uxengine:
Before:
screen shot 2017-12-30 at 14 13 53

After:
screen shot 2017-12-30 at 14 14 14

### DIFF
--- a/src/components/Figure/utils.js
+++ b/src/components/Figure/utils.js
@@ -6,7 +6,7 @@ import {
   MAX_WIDTH_MOBILE
 } from '../Center'
 
-export const getResizedSrcs = (src, displayWidth, fillMaxWidth = true) => {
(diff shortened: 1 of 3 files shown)


src/components/Figure/utils.js
```diff
@@ -6,7 +6,7 @@ import {
   MAX_WIDTH_MOBILE
 } from '../Center'
 
-export const getResizedSrcs = (src, displayWidth) => {
+export const getResizedSrcs = (src, displayWidth, fillMaxWidth = true) => {
```
tpreusse: Nit: I'd call it `enforce` or `set` `MaxWidth`

uxengine: My good ol' Google slang, fill objects all day long :)

PR #73, opened 2018-01-02
fix(EditorialNewsletter): Add link to web version in email header and footer

uxengine:
Mailchimp preview:
screen shot 2018-01-02 at 16 40 32

### DIFF
--- a/src/templates/EditorialNewsletter/email/Header.js
+++ b/src/templates/EditorialNewsletter/email/Header.js
@@ -14,7 +14,6 @@ export default withMeta(({ meta }) => {
         <a
           href={`https://www.republik.ch/${slug}`}
           title="Im Web lesen"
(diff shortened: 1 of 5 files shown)


uxengine:
FYI, also linked the header logo to the web view now.

src/templates/EditorialNewsletter/email/Header.js
```diff
@@ -21,10 +26,10 @@ export default () => {
               maxWidth: '100% !important'
             }}
             width="180"
-            alt="REPUBLIK"
+            alt="Im Web lesen"
```
tpreusse: alt should still be republik. title should suffice

uxengine:
Done. On the container, I still have to use node.meta and pipe that through to the header component.

PR #76, opened 2018-01-03
fix(TeaserFeed): Fall back to publish date if a feed item has no credits

uxengine:
Makes sure published newsletters appear with a date in the feed.

Preview:
screen shot 2018-01-03 at 16 42 15


tpreusse:
merged via https://github.com/orbiting/styleguide/pull/77

PR #81, opened 2018-01-10
Ux love

lukasbuenger:
Depends on orbiting/publikator-frontend#108

##### Control flow editorOptions
Values:
- isStatic, Boolean, default false. Whether the component should behave as a static field.
- afterType, String, default undefined. Component to insert if Enter is pressed.
- insertAfterType, String. Component that should get used as insertion target when Enter is pressed. Default is the root document.

Currently enabled on the modules paragraph, headline and figureCaption. If no afterType is present, Enter performs just a selection change to the next block. afterTypes work best on the last static field of a composed component.

##### insertTypes
Simple whitelist of elements for which the insert button should get enabled. Currently enabled on image, imageGroup, quote and infobox


src/templates/Article/index.js
```diff
@@ -92,24 +83,46 @@ const paragraph = {
...
+    }
+  },
+  editorModule: 'figureImage',
+  isVoid: true
+}
+
+const figureByLine = {
+  matchMdast: matchType('emphasis'),
+  component: FigureByline,
+  // ToDo: inline module with placeholder
```
tpreusse: rm todo

src/templates/Article/index.js
```diff
@@ -123,49 +136,194 @@ const figure = {
...
+  ]
+}
 
-        return {
-          ...FigureImage.utils.getResizedSrcs(
-            src,
-            displayWidth
-          ),
-          alt: node.children[0].alt
+const floatFigure = {
```
tpreusse: would prefer the naming `centerFigure`—was trigging to figure out why the figure floats ;)

PR #87, opened 2018-01-17
fix(template): fix the mdast matcher for quote body

lukasbuenger:


src/templates/Article/index.js
```diff
@@ -295,10 +295,7 @@ const pullQuote = {
   rules: [
     figure,
     {
-      matchMdast: (node, index, parent) =>
-        matchParagraph(node) &&
-        (index === 0 ||
-          !matchLast(node, index, parent)),
+      matchMdast: (node, index, parent) => matchParagraph(node) && index < 2,
```
tpreusse: A quote without a figure would now match the source here. Lucky we don't have any quotes with sources and without images live yet. Additionally this assumes there is only ever one paragraph. This is currently in line with the behaviour in publikator, however I always assumed it could be more.

tpreusse: Fixed in https://github.com/orbiting/styleguide/commit/35f4a1fd5038ab7bef86e3f22d8d1e381c9eaa50

PR #90, opened 2018-01-23
fix(Figure): 5px top-margin for figure caption

uxengine:
This was introduced by image placeholders (display: block), while unwrapped image's (display: inline) captions still have some default top spacing inherited from line-height. To reset spacing for both cases, we're also using `display: block` on normal images now and explicitly set the top margin on the caption.

Before:
screen shot 2018-01-23 at 18 05 34

After:
screen shot 2018-01-23 at 18 05 53

### DIFF
--- a/src/components/Figure/index.js
+++ b/src/components/Figure/index.js
@@ -15,7 +15,6 @@ export { default as FigureByline } from './Byline'
 
 const styles = {
   figure: css({
-    lineHeight: 0,
     margin: 0,
     marginBottom: 15,
     padding: 0,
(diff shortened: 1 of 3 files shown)


src/components/Figure/index.js
```diff
@@ -15,6 +15,7 @@ export { default as FigureByline } from './Byline'
 
 const styles = {
   figure: css({
+    lineHeight: 0,
```
tpreusse: could this be moved to figure image? Because line height are inherited it can be a bit dangerous in element that do have text children

uxengine: `line-height:0` on the image and image wrappers doesn't help, but I've now switched to resetting the image to `display: block` instead of using line-height. So both the "native" and the placeholder images are `display: block` now.

PR #93, opened 2018-01-30
fix(EditorialNewsletter): Don't max-width cover image

uxengine:
Also delete unused cover/center component.

Buggy rendering:
screen shot 2018-01-29 at 16 34 17


uxengine:
Hang on, there is still something fishy going on. I'll let you know once this works as intended.

uxengine:
@tpreusse ready to review now.

PR #95, opened 2018-02-01
feat: optional centerAppend for article schema

uxengine:


uxengine:
Ignore, not needed anymore.

PR #99, opened 2018-02-06
feat(FrontTile): add image-only use case, e.g. for cartoons

uxengine:
Allows to display a full-width or full-height image in a front tile.

Previews:
https://r-styleguide-pr-99.herokuapp.com/teaserfronttile#image-only
https://r-styleguide-pr-99.herokuapp.com/templates/front

In Publikator, all we'd need is an "imageOnly" checkbox for tiles.

Depends on https://github.com/orbiting/publikator-frontend/pull/131

### DIFF
--- a/src/components/TeaserFront/Tile.js
+++ b/src/components/TeaserFront/Tile.js
@@ -65,7 +65,7 @@ const styles = {
       margin: '0 auto 60px auto'
     }
   }),
-  imageOnlyContainer: css({
+  onlyImageContainer: css({
     margin: '0 auto',
     fontSize: 0
   }),
@@ -78,6 +78,11 @@ const styles = {
     [tUp]: {
       ...sizeLarge
     }
+  }),
+  onlyImage: css({
+    minWidth: '100px',
+    maxHeight: '100% !important',
+    maxWidth: '100% !important'
   }),
   row: css({
     margin: 0,
@@ -167,7 +172,7 @@ const Tile = ({
   bgColor,
   align,
   aboveTheFold,
-  imageOnly
+  onlyImage
 }) => {
   const background = bgColor || ''
   const justifyContent =
@@ -182,13 +187,9 @@ const Tile = ({
     cursor: onClick ? 'pointer' : 'default',
     justifyContent
   }
-  if (imageOnly) {
-    containerStyle = {
-      ...containerStyle,
-      padding: 0
-    }
+  if (onlyImage) {
+    containerStyle.padding = 0
   }
-  const imageStyle = imageOnly ? {maxHeight: '100%', maxWidth: '100%'} : {}
 
   return (
     <div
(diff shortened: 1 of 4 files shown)


src/components/TeaserFront/Tile.js
```diff
@@ -162,32 +177,40 @@ const Tile = ({ children, attributes, image, alt, onClick, color, bgColor, align
...
+      style={containerStyle}
       className='tile'
     >
       {imageProps && (
-        <div {...styles.imageContainer}>
+        <div {...(imageOnly ? styles.imageOnlyContainer : styles.imageContainer)}>
           <LazyLoad visible={aboveTheFold}>
             <img src={imageProps.src} srcSet={imageProps.srcSet} alt={alt}
-              {...styles.image} />
+              {...styles.image} style={imageStyle}/>
```
tpreusse: Use an `imageOnly` css style here too. Feels odd to use a different pattern for the image container vs image.

src/components/TeaserFront/Tile.js
```diff
@@ -162,32 +177,40 @@ const Tile = ({ children, attributes, image, alt, onClick, color, bgColor, align
...
+    cursor: onClick ? 'pointer' : 'default',
+    justifyContent
+  }
+  if (imageOnly) {
+    containerStyle = {
+      ...containerStyle,
+      padding: 0
+    }
+  }
+  const imageStyle = imageOnly ? {maxHeight: '100%', maxWidth: '100%'} : {}
```
tpreusse: mv to styles and use the same pattern as with `imageOnlyContainer` vs. `imageContainer`

src/components/TeaserFront/Tile.js
```diff
@@ -162,32 +177,40 @@ const Tile = ({ children, attributes, image, alt, onClick, color, bgColor, align
     IMAGE_SIZE.large,
     false
   )
+  let containerStyle = {
+    background,
+    cursor: onClick ? 'pointer' : 'default',
+    justifyContent
+  }
+  if (imageOnly) {
+    containerStyle = {
```
tpreusse: `containerStyle.padding = 0` would do too

src/components/TeaserFront/Tile.js
```diff
@@ -153,7 +157,18 @@ TeaserFrontTileRow.defaultProps = {
...
+  children,
+  attributes,
+  image,
+  alt,
+  onClick,
+  color,
+  bgColor,
+  align,
+  aboveTheFold,
+  imageOnly
```
tpreusse: Tile already has an `showImage` option in the editor—`onlyImage` would therefore be more consistent

src/components/TeaserFront/Tile.js
uxengine: The reason for this pattern was specificity (style attribute is more specific than the tilegroup's parent selectors affecting maxwidth/maxheight). I'm using a consistent pattern now, at the cost of introducing !importants.

PR #105, opened 2018-02-13
feat(VideoPlayer): add fullscreen mode

uxengine:
Try out: https://r-styleguide-pr-105.herokuapp.com/videoplayer

- New fullscreen icon appears if browser supports fullscreen API
- In fullscreen mode, native controls are enabled

Chrome inline:
screen shot 2018-02-13 at 18 49 27

Chrome fullscreen:
screen shot 2018-02-13 at 18 40 14

Firefox fullscreen:
screen shot 2018-02-13 at 18 47 13

Safari fullscreen:
screen shot 2018-02-13 at 18 51 21

MS Edge fullscreen:
screen shot 2018-02-13 at 18 31 09

IE11 fullscreen:
screen shot 2018-02-13 at 16 11 43

iOS shows no fullscreen icon since API is not supported and it's playing fullscreen anyway.
screen shot 2018-02-13 at 18 38 44

### DIFF
--- a/src/components/VideoPlayer/Fullscreen.js
+++ b/src/components/VideoPlayer/Fullscreen.js
@@ -44,7 +44,7 @@ const apiSurfaces = [
   ]
 ]
 
-const fullscreenFunction = () => {
+const getFullscreenApi = () => {
   for (let i = 0; i < apiSurfaces.length; i++) {
     let apiSurface = apiSurfaces[i]
     if (!!document[apiSurface[1]]) {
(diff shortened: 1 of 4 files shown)


src/components/VideoPlayer/Fullscreen.js
```diff
@@ -0,0 +1,90 @@
...
+      for (let j = 0; j < apiSurface.length; j++) {
+        api[apiSurfaces[0][j]] = apiSurface[j]
+      }
+      return api
+    }
+  }
+  return null
+}
+
+const Fullscreen = () => {
```
tpreusse: capital letter should only be used for components (react spec) and classes. This is a class like structure but not really one. Currently this also feels more like boilerplate wrapper than actual help. You'd most likely save 20 or more lines by not abstracting it here and directly calling `document.addEventListener`, `document.removeEventListener`, `elem[api.requestFullscreen]()` and `document[api.fullscreenElement]` in the component.

An alternative would be to use a different pattern. E.g.
```js
const setupFullscreen = ({onChange, onError}) => {
  const api = getFullscreenApi()
  if (!api) {
    return
  }

  document.addEventListener(api.fullscreenchange, onChange, false)
  document.addEventListener(api.fullscreenerror, onError, false)

  return {
    request (elem) {
      const subject = elem || document.documentElement
      subject[api.requestFullscreen]()
    }
    element () {
      return document[api.fullscreenElement]
    }
    dispose () {
      document.removeEventListener(api.fullscreenchange, onChange, false)
      document.removeEventListener(api.fullscreenerror, onError, false)
    }
  }
}
```

Then just do some like this in the constructor:
```js
this.fullscreen = setupFullscreen({
  onChange () {
    this.setState({isFullscreen: this.fullscreen.element() === this.video}))
  },
  onError () {
     this.setState({isFullscreen: false})
  }
})
```
and `this.fullscreen && this.fullscreen.dispose()` in `componentWillUnmount`.

Notice the check that it is actually the video element that from the component that goes full screen and not another element. What is unclear to me: is setting `isFullscreen: false` onError truly necessary? Does the error not exclusively occur on request, before `this.fullscreen.element() === this.video` ever becomes true?

uxengine:
All comments addressed.

Correct, setting isFullscreen: false onError is not necessary since fullscreenchange and fullscreenerror events are mutually exclusive according to https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen:
> If permission to enter full screen mode is granted, the document will receive a fullscreenchange event to let it know that it's now in full screen mode. If permission is denied, the document receives a fullscreenerror event instead.

PR #110, opened 2018-02-19
fix(AudioPlayer): more useful UX when source failed loading

uxengine:
Instead of keeping the loading spinner around endlessly, provide a meaningful message and a link to retry.

screen shot 2018-02-19 at 17 56 47

### DIFF
--- a/src/components/AudioPlayer/Player.js
+++ b/src/components/AudioPlayer/Player.js
@@ -277,14 +277,14 @@ class AudioPlayer extends Component {
     this.setInstanceState = state => {
       this.setState(state)
     }
-  }
-  toggle() {
-    const { audio } = this
-    if (audio) {
-      if (audio.paused || audio.ended) {
-        this.play()
-      } else {
-        this.pause()
+    this.toggle = () => {
+      const { audio } = this
+      if (audio) {
+        if (audio.paused || audio.ended) {
+          this.play()
+        } else {
+          this.pause()
+        }
       }
     }
   }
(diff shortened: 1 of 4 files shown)


src/components/AudioPlayer/Player.js
```diff
@@ -319,9 +362,9 @@ class AudioPlayer extends Component {
           onLoadedMetadata={this.onLoadedMetaData}
           crossOrigin="anonymous"
         >
-          {src.mp3 && <source src={src.mp3} type="audio/mpeg" />}
-          {src.aac && <source src={src.aac} type="audio/mp4" />}
-          {src.ogg && <source src={src.ogg} type="audio/ogg" />}
+          {src.mp3 && <source src={src.mp3} type="audio/mpeg" onError={(e) => this.onSourceError(e)} />}
```
tpreusse: `this.onSourceError` is bound, no need for the extra function: `onError={this.onSourceError}`—`this.onSourceError = () => {}` arrow function in the constructor are always bound. 
[constructor arrow fn vs «method» fn](http://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAEAqCmEAu0DeAoa1gHsB2yATgK7BI6HQAUAlGpltEgBYCWEAdGIYTgO7QAvNVx4k8AB5I6ggHz1GjURBwh4HEDgDmVAOTdefXQBps-cVNMt20VgSRg8weDgBmcRNIZYAvgz9YALbwLDgAJiLmktIKSvgqahraesGhYSZmYtFWbDB2yI7Obh7INP4MhPB4YfCUtLFY1pwG_HpNumWKTRypzOFtuR3e0BAhsKzBOMRIVN0tfKYAjAAMy52Mo0jjk9OzuT0hfWFLq-t-fujoyijiyELQePACCMi0l7dIHJXVtW8fXDxWrp2mV_r1-sDBmUgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=stage-3%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2)
_there are other solutions, like `.bind` and class fields, also see [react docs](https://reactjs.org/docs/handling-events.html), I try to avoid `.bind` and use constructor arrow fns when I want to avoid the overhead of `() => {}`—creating a new function on every render_

you should also move `this.toggle` to the constructor, since it recreates a fn on very render (on every time sync)

PR #113, opened 2018-02-22
Dossier

lukasbuenger:
https://github.com/orbiting/styleguide/compare/dossier?expand=1#diff-160b35e79868a99a2467231055f7f017L542
This is, where I tried to directly cut off the link into a block section.


src/templates/Article/teasers.js
```diff
@@ -187,10 +191,9 @@ const createTeasers = ({
       </TeaserFrontTileRow>
     },
     // TMP: Disabled until editor integration
```
tpreusse: rm

src/templates/Article/teasers.js
```diff
@@ -208,10 +211,11 @@ const createTeasers = ({
       props: node => node.data,
       // TMP: Disabled until editor integration
```
tpreusse: rm

src/templates/Front/index.js
```diff
@@ -529,39 +529,45 @@ const createSchema = ({
       </TeaserFrontDossier>
     },
     // TMP: Disabled until editor integration
```
tpreusse: rm

PR #114, opened 2018-02-23
fix: fix dossier teaser behavior on front. remove dnd key as it is no…

lukasbuenger:
…w longer needed


tpreusse:
custom message
screen shot 2018-02-23 at 18 08 25
default message
screen shot 2018-02-23 at 17 59 44

PR #118, opened 2018-02-26
fix: dossier template docs

uxengine:
mdast rendering is currently broken on
https://styleguide.republik.ch/templates/dossier
due to outdated markdown.


uxengine:
No probs, I actually wanted to check the test-deploy while reviewing but then things took too long. So I'm basically fixing my own incomplete review here :)

PR #122, opened 2018-03-20
feat: gallery

thommymo:


tpreusse:
@garamond I'm fine with adding story book in `republik-frontend` if you want to use it. But adding it here seems wrong—to unclear when to do a story and when a catalog page/snippet. Please remove again and covert you stories to catalog pages.

src/components/Figure/Image.js
```diff
@@ -25,18 +25,21 @@ class Image extends Component {
       attributes = {},
       maxWidth,
       size: sizeProp,
-      aboveTheFold
+      aboveTheFold,
+      enableGallery
     } = this.props
 
+    const onClick = (enableGallery && this.context.toggleGallery) || function(){}
```
uxengine: Thomas would probably prefer the ES6 no-op function:
`... || () => {}`

src/components/Figure/Image.js
```diff
@@ -61,6 +64,10 @@ Image.propTypes = {
   aboveTheFold: PropTypes.bool
 }
 
+Image.contextTypes = {
+  toggleGallery: PropTypes.func
```
uxengine: Cool, I didn't know about the context feature in React. I assume it would be hard to pipe `toggleGallery` down to the component through the schema otherwise.

At first it looked a bit fishy to me to have the styleguide component depend on the context of the frontend parent component. But since it's optional, it's probably not that fishy. Would you see any other caveats from your experience?

src/components/Gallery/Gallery.js
```diff
@@ -0,0 +1,345 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { css, merge } from 'glamor'
+import zIndex from '../../theme/zIndex'
+import { onlyS, lUp } from '../../theme/mediaQueries'
+import debounce from "lodash.debounce";
```
uxengine: Here and two more imports below:
- single quotes
- no semicolons

src/components/Gallery/Gallery.js
```diff
@@ -0,0 +1,345 @@
...
+          <NavOverlay
+            handleClickLeft={this.handleClickLeft}
+            handleClickRight={this.handleClickRight}
+            handleClick={this.toggleFocus}
+            onClose={onClose}
+          />
+          <div {...styles.gallery}>
+            <div {...styles.header} style={{ opacity: focus ? 0 : 1 }}>
+              <div {...styles.counter}>
+                <span>{index+1}/{total}</span>
```
uxengine: `span` really needed?

src/components/Gallery/Gallery.js
```diff
@@ -0,0 +1,345 @@
...
+              </div>
+              <div {...styles.close} onClick={onClose}>
+                <Close size={24} />
+              </div>
+            </div>
+            <div {...styles.body}>
+              <div {...merge(
```
uxengine: Since we're trying to avoid merge at render time, and there's no overlapping style properties as far as I see, would something like this also work?

```
<div
  {...styles.mediaItem}
  {...(exitLeft && styles.exitLeft)}
  {...(exitRight && styles.exitRight)}
  {...(closing && styles.closing)}
>
```

src/components/Gallery/Gallery.js
```diff
@@ -0,0 +1,345 @@
...
+}
+
+Gallery.propTypes = {
+  items: PropTypes.arrayOf(PropTypes.shape({
+    src: PropTypes.string.isRequired,
+    alt: PropTypes.string.isRequired,
+    caption: PropTypes.string,
+    credit: PropTypes.string
+  })),
+  startItem: PropTypes.string,
```
uxengine: Maybe call it `startItemSrc` to make clear what to expect here.

src/components/Gallery/Gallery.js
```diff
@@ -0,0 +1,345 @@
...
+  navButton: css({
+    color: '#fff',
+    display: 'none',
+    [lUp]: {
+      display: 'block',
+    }
+  }),
+}
+
+function removeQuery(url) {
```
uxengine: Wearing the ES6 hat again:
`const removeQuery = url => (url || '').split('?')[0]`

src/components/Gallery/Gallery.js
```diff
@@ -0,0 +1,345 @@
...
+  componentDidMount() {
+    this.ref && this.ref.focus()
+  }
+  render() {
+    const { handleClickLeft, handleClickRight, onClose, handleClick } = this.props
+    return (
+      <div
+        ref={this.setRef}
+        {...styles.nav}
+        onKeyDown={e => this.handleKeyUp(e, handleClickLeft, handleClickRight, onClose)}
```
uxengine: I think you don't need to pass `handleClickLeft, handleClickRight, onClose` to `handleKeyUp` here. They're not used from the arguments list there, but from props.

src/components/Gallery/Gallery.js
```diff
@@ -0,0 +1,345 @@
...
+}
+
+const styles = {
+  wrapper: css({
+    position: 'fixed',
+    top:0,
+    right: 0,
+    bottom:0,
+    left:0,  
+    zIndex: 100
```
uxengine: Move to theme/zIndex.js and then e.g. `zIndex.gallery`

src/components/Figure/Image.js
garamond: Prior to React v16.3, context used to be an "undocumented feature" to pass data through the component tree with out explicit props – the [new, official Context API](https://reactjs.org/docs/context.html) will be even nicer and easier to use.

I don't think having context provider and consumer in different projects is a problem – it works just like props, but behind the scenes. However, the need to use context is usually a hint that the design is flawed – but in this case I see no other way to distribute the global handler through the schema.

src/components/Gallery/Gallery.js
garamond: Excellent idea, thanks!

src/components/Gallery/Gallery.js
garamond: I agree that the ES6 syntax looks nicer – the reason I still use `function` sometimes (in reality, whenever I'm not writing an inline lambda) is that the name of a named function appears in stack traces whereas the name of the variable a lambda is bound to won't.

That said, I agree that we should agree on a style…

src/components/Gallery/Gallery.js
```diff
@@ -0,0 +1,245 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { css } from 'glamor'
+import { onlyS } from '../../theme/mediaQueries'
+import debounce from "lodash.debounce";
```
uxengine: I still think it would be nice to consistently use single quotes and no semicolons in our imports (here and below).

garamond: Sorry, somehow that change didn't make it in my last commit…

PR #123, opened 2018-03-22
fix(AudioPlayer): resolve TimeRanges object without using keys()

uxengine:
Currently, on IE11 the AudioPlayer explodes when playback starts with this error message:
screen shot 2018-03-22 at 11 41 10

The error message suggests that IE11 has issues with Array.prototype.keys() at least in republik-frontend context -- which in fact is executed only when playing starts.

Currently I can't test this on IE11 locally, but I would like to check this speculative fix on .love.

TimeRanges interface: https://developer.mozilla.org/en-US/docs/Web/API/TimeRanges

### DIFF
--- a/src/components/AudioPlayer/Player.js
+++ b/src/components/AudioPlayer/Player.js
@@ -419,6 +419,13 @@ class AudioPlayer extends Component {
       right: timePosition === 'right' ? rightIconsWidth + 10 : 'auto'
     }
 
+    let timeRanges = []
+    if (buffered && !!buffered.length) {
+      for (let i = 0; i < buffered.length; i++) {
+        timeRanges.push({start: buffered.start(i), end: buffered.end(i)})
+      }
+    }
+
     return (
       <div {...merge(styles.wrapper, breakoutStyles[size])}
         ref={this.containerRef}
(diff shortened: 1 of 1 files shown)


src/components/AudioPlayer/Player.js
```diff
@@ -508,15 +515,13 @@ class AudioPlayer extends Component {
           />
           <div {...styles.buffer}>
             {this.audio &&
-              buffered &&
-              !!buffered.length &&
-              Array.from(Array(buffered.length).keys()).map(index => (
```
tpreusse: Another alternative: [d3.range](https://github.com/d3/d3-array#range)—`range(buffered.length)`

src/components/AudioPlayer/Player.js
```diff
@@ -419,6 +419,13 @@ class AudioPlayer extends Component {
       right: timePosition === 'right' ? rightIconsWidth + 10 : 'auto'
     }
 
+    let timeRanges = []
+    if (buffered && !!buffered.length) {
```
tpreusse: unnecessary boolean forcing (`!!`) only needed in a react context because `false` mean no child and e.g. `0` (falsy) is rendered as text `0`

PR #127, opened 2018-04-24
feat: Comments mdast

uxengine:
# Web rendering in republik-frontend:
screen shot 2018-04-24 at 19 19 57

# Email rendering
screen shot 2018-04-26 at 17 03 26

### DIFF
--- a/src/components/Comment/Comment.js
+++ b/src/components/Comment/Comment.js
@@ -2,7 +2,6 @@ import React from 'react'
 import { css } from 'glamor'
 import { renderMdast } from 'mdast-react-render'
 
-import { serifRegular14, serifRegular16 } from '../Typography/styles'
 import { Label } from '../Typography'
 
 import { mUp } from '../../theme/mediaQueries'
@@ -35,11 +34,7 @@ const styles = {
     marginTop: 12
   }),
   body: css({
-    margin: `12px 0 12px ${profilePictureSize + profilePictureMargin}px`,
-    ...serifRegular14,
-    [mUp]: {
-      ...serifRegular16
-    }
+    margin: `12px 0 12px ${profilePictureSize + profilePictureMargin}px`
   })
 }
 
--- a/src/components/CommentBody/web/Blockquote.js
+++ b/src/components/CommentBody/web/Blockquote.js
@@ -12,7 +12,6 @@ const styles = {
     backgroundColor: '#f7f7f7',
     margin: 0,
     padding: '0 15px 12px 15px',
-    fontSize: '15px',
     ...serifRegular14,
     [mUp]: {
       ...serifRegular16,


src/templates/Comment/schema.js
```diff
@@ -0,0 +1,173 @@
...
+
+  const heading = {
+    matchMdast: matchType('heading'),
+    component: Heading
+  }
+
+  const paragraph = {
+    matchMdast: matchParagraph,
+    component: ({ children }) => <Paragraph>{children}</Paragraph>,
+    rules: [
```
tpreusse: `rules: globalInlines`

src/templates/Comment/schema.js
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  const blockQuote = {
+    matchMdast: matchType('blockquote'),
+    component: BlockQuote,
+    rules: [
+      ...globalInlines,
+      {
+        matchMdast: matchParagraph,
+        component: ({ children }) => children
```
tpreusse: What about multi paragraph quotes?
> Paragraph one
> on two lines
>
> Paragraph two on another line

```md
> Paragraph one
> on two lines
>
> Paragraph two on another line
```

```json
{
  "type": "root",
  "children": [
    {
      "type": "blockquote",
      "children": [
        {
          "type": "paragraph",
          "children": [
            {
              "type": "text",
              "value": "Paragraph one\non two lines"
            }
          ]
        },
        {
          "type": "paragraph",
          "children": [
            {
              "type": "text",
              "value": "Paragraph two on another line"
            }
          ]
        }
      ]
    }
  ]
}
```

src/templates/Comment/schema.js
```diff
@@ -0,0 +1,173 @@
...
+    matchMdast: matchType('thematicBreak'),
+    component: HR
+  }
+
+  const blockLevelHtml = {
+    matchMdast: matchType('html'),
+    props: node => ({
+        value: node.value
+      }),
+      component: ({value}) => <Paragraph>{value}</Paragraph>
```
tpreusse: fix indentation

src/components/CommentBody/email/List.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+import { fontFamilies } from '../../../theme/fonts'
```
tpreusse: unused

src/components/Comment/Comment.js
```diff
@@ -38,6 +43,12 @@ const styles = {
   })
 }
 
+export const CommentBody = ({children, style}) => (
```
tpreusse: I think it would be better to have a comment root component in the mdast schema with the font sizes (serifRegular14/serifRegular16) and not expose this here—only have an wrapper with the profile pic margin here.

src/components/BlockQuote/BlockQuote.js
```diff
@@ -17,11 +17,12 @@ const styles = {
   })
 }
 
-const BlockQuote = ({ children, attributes }) => {
+const BlockQuote = ({ children, attributes, style }) => {
```
tpreusse: Overwriting styles like this make maintenance hard. And when overwriting media query behaviour things get hard to understand.

IMO just write a wrapper with it's own margin in `components/CommentBody/web/Blockquote.js` everything else is not needed in the comment case anyhow.

src/components/BlockQuote/BlockQuoteParagraph.js
```diff
@@ -25,11 +25,12 @@ const styles = {
   })
 }
 
-const BlockQuoteParagraph = ({ children, attributes }) => {
+const BlockQuoteParagraph = ({ children, attributes, style }) => {
```
tpreusse: same as above: write a comment specific block quote paragraph in `components/CommentBody/web/Blockquote.js`. Overwriting media query styles is too hard to debug. DRY benefit is small here.

src/templates/Comment/schema.js
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  return {
+    rules: [
+      {
+        matchMdast: matchType('root'),
+        component: ({ children }) => children,
```
tpreusse: New wrapper with media query font sizes would go here. Configurable for the email case I guess.

src/lib.js
```diff
@@ -93,6 +93,14 @@ export {
   CommentHeader,
   CommentActions
 } from './components/Comment'
+export {
```
tpreusse: IMO skip exposing individual elements unless truly needed. This way we can refactor how the schema is implemented without breaking a «public» API.

src/components/CommentBody/web/Blockquote.js
```diff
@@ -0,0 +1,40 @@
...
+
+const styles = {
+  blockquote: css({
+    margin: '20px auto'
+  }),
+  paragraph: css({
+    backgroundColor: '#f7f7f7',
+    margin: 0,
+    padding: '0 15px 12px 15px',
+    fontSize: '15px',
```
tpreusse: will be overwritten right? rm

src/components/CommentBody/web/List.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+import { List, ListItem as CommentListItem} from '../../List'
+
+export const ListItem = ({ children }) => (
+  <CommentListItem style={{
+    fontSize: 'inherit',
```
tpreusse: is this needed? Actual text is always in a child paragraph right?

uxengine: It's needed for the list style items (1., 2., 3.) which are :before items to match the font cut, otherwise we'd end up with this mismatch:
screen shot 2018-04-30 at 17 35 57

uxengine:
FYI, commit  d207609 was a bit WIP, but I wanted to unblock Patte's backend work.

Significant changes:
- We're now supporting multiline blockquotes (see screenshots below)
- Also added support to preserve inline [...] which would otherwise be considered a linkReference with missing mdast support.
- Removed all newly introduced `style` properties to keep things clean
- Added root container component both for web and email

### Web 
screen shot 2018-04-30 at 17 07 51
### Email
screen shot 2018-04-30 at 17 08 10

src/lib.js
uxengine: I agree, but I'm using them all on the cheat sheet page.

PR #129, opened 2018-05-02
feat(Comment): code block, ellipsized hrefs, linkReference and definition (i.e. footnote)

uxengine:
Note that shortcut link references like `[...]` are still left as is to allow e.g. proper citation, while full link references like`[Label][1]` are converted to `Label [1]`, which can/should be combined with definitions (footnotes).

Sorry, some commit artefacts from late merging.

## Web
screen shot 2018-05-02 at 19 40 01

## Email
screen shot 2018-05-02 at 19 40 23


src/templates/Comment/schema.js
```diff
@@ -130,6 +151,16 @@ const createCommentSchema = ({
     ]
   }
 
+  const blockCode = {
+    matchMdast: matchType('code'),
+    props: node => ({
+      value: node.value
+    }),
+    component: ({value}) => <BlockCode>
+      {value.split('\n').map((line, i) => <Fragment key={i}>{line}<br /></Fragment>)}
```
tpreusse: Use `white-space: pre;` or a `<pre>` element instead

PR #132, opened 2018-05-08
feat(Field): search and reset

uxengine:


uxengine:
@tpreusse Done, see last example on https://r-styleguide-pr-132.herokuapp.com/forms#fields

src/components/Form/Field.js
```diff
@@ -111,6 +114,12 @@ const arrowDownStyle = css({
   top: lineHeight + fieldHeight / 2 - 3,
   cursor: 'pointer'
 })
+const iconWrapperStyle = css({
+  position: 'absolute',
+  right: 3,
+  top: fieldHeight - (fieldHeight * .75 / 2),
```
tpreusse: better definition: `lineHeight + 5` or `lineHeight + fieldHeight  * 0.25 / 2`

`lineHeight` is the label hight (top padding of the `<label>`, but absolute ignores padding).

src/components/Form/Field.js
```diff
@@ -111,6 +114,12 @@ const arrowDownStyle = css({
...
+  cursor: 'pointer'
```
tpreusse: IMO rm, the icon element should decide if it is clickable and what cursor it should have, but can be overwritten either way

src/components/Form/Field.js
tpreusse: (currently this only works because `lineHeight * 2 === fieldHeight`)

PR #138, opened 2018-06-04
fix(frontimage): bottomleft/right positioning of text should be relative to bottom (instead of 50%)

uxengine:
Typical issue in production with longer leads:
screen shot 2018-06-04 at 10 58 48

With the fix, text may grow upwards:
screen shot 2018-06-04 at 11 28 53


uxengine:
@tpreusse  Do you mean overflow:hidden specifically for the lead element or on the teaser container? In any case, I'll handle that separately.

tpreusse:
@uxengine teaser container

Issue #139, opened 2018-06-04
FrontImage: Hide overflow on teaser container


uxengine:
We had some issues with long (unbreakable) words.

overflow: hidden; needs a clear bounding box. This can be accomplished by setting (left, top, right and bottom) or afaik combo of (bottom, left, right, max-height) should work too.

PR #140, opened 2018-06-11
fix(FrontImage): hide overflow both for unpositioned (mobile) and positioned (desktop) layout

uxengine:
Fixes https://github.com/orbiting/styleguide/issues/139

### DIFF
--- a/src/components/TeaserFront/Image.js
+++ b/src/components/TeaserFront/Image.js
@@ -15,6 +15,7 @@ const styles = {
     }
   }),
   textContainer: css({
+    overflow: 'hidden',
     padding: '15px 15px 40px 15px',
     [mUp]: {
       padding: '40px 15% 70px 15%'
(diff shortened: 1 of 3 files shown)


src/components/TeaserFront/Image.js
```diff
@@ -15,6 +15,7 @@ const styles = {
     }
   }),
   textContainer: css({
+    overflow: 'hidden',
```
tpreusse: Does this have any effect? AFAIK this does nothing, if so please remove. If not add a comment in which case this helps.

uxengine: Yes, it hides unpositioned content in the stacked case on mobile.

Without this fix:
screen shot 2018-06-11 at 17 12 48

With this fix:
screen shot 2018-06-11 at 17 13 31

PR #141, opened 2018-06-14
feat(DynamicComponent): load dynamic, story specific components in articles

tpreusse:
First version, only client side dynamic rendering.

An dynamic component can be written like this:
https://gist.github.com/tpreusse/4e1c8050cc6c8f4392a65ae98889810d

JS loading with dependency injection via `d3-require` (using a custom `d3.requireFrom` and alias `react`, `glamor` and `@project-r/styleguide`).

This includes an stripped down `d3-require`—`src/components/DynamicComponent/require.js`. I had to remove the default resolver which uses template literals and async/await. Since we don't use the default resolver anyhow this seems like the easiest and even most size efficient solution. You don't have to review that `d3-require`, it's pretty narly code to read though ;)

Docs:
- [DynamicComponent](https://r-styleguide-pr-141.herokuapp.com/dynamic-component)
- [In Article Template](https://r-styleguide-pr-141.herokuapp.com/templates/article#dynamic-components)


src/components/DynamicComponent/index.js
```diff
@@ -0,0 +1,89 @@
...
+      {...loader}
+      loading />
+  }
+}
+
+DynamicComponent.propTypes = {
+  src: PropTypes.string.isRequired,
+  html: PropTypes.string,
+  props: PropTypes.object,
+  loader: PropTypes.object,
```
uxengine: loaderProps?

tpreusse: initial had `loaderProps` but felt repetitive: `this.props.loaderProps`—it feels given that props are props. But I have the same difficulty with `this.props.props`. Maybe it should be long form? `componentProps` and `loaderProps`?

PR #143, opened 2018-06-18
feat(DynamicComponent): edge-to-edge and raw (without figure wrapper)

tpreusse:
Support Edge-to-Edge dynamic components:
screen shot 2018-06-18 at 17 49 36

and raw mode without figure wrapping.

### DIFF
--- a/src/templates/Article/docs.md
+++ b/src/templates/Article/docs.md
@@ -769,18 +769,26 @@ Zweitens: ich habe erklärt mit diese zwei Spieler: nach Dortmund brauchen viell
 
 ## Dynamic Components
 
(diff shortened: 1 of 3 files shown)


src/templates/Article/docs.md
```diff
@@ -791,6 +799,47 @@ Zweitens: ich habe erklärt mit diese zwei Spieler: nach Dortmund brauchen viell
...
 
+Denn esse est percipi - Sein ist wahrgenommen werden.
+
+<hr /></section>
+`}</Markdown>
+```
+
+### Edge-to-Edge
+
+Simpy unwrap from center
```
uxengine: Typo: Simpy -> Simply
(also: maybe `detach` instead of `unwrap`)

tpreusse: Thx for the review. `unwrap` is a slate term so leaving it at that.

PR #146, opened 2018-07-09
fix(CommentsTeaser): graciously fall back on span when mdast node has…

lukasbuenger:
… no renderer

Test markdown snippet
```md
ich denke

- also bin ich
   - eingerückt
```

Before:
image

Now:
image

### DIFF
--- a/src/components/CommentTeaser/CommentTeaser.js
+++ b/src/components/CommentTeaser/CommentTeaser.js
@@ -61,7 +61,11 @@ export const CommentTeaser = ({
         lineClamp ? merge(styles.clamp, { WebkitLineClamp: lineClamp }) : {}
       )}
     >
-      {renderMdast(content, schema)}
+      {renderMdast(
+        content,
+        schema,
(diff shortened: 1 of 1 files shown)


src/components/CommentTeaser/CommentTeaser.js
```diff
@@ -61,7 +61,11 @@ export const CommentTeaser = ({
         lineClamp ? merge(styles.clamp, { WebkitLineClamp: lineClamp }) : {}
       )}
     >
-      {renderMdast(content, schema)}
+      {renderMdast(
+        content,
+        schema,
+        { MissingNode: ({ children }) => <span>{children}</span> }
```
tpreusse: We should use the same `MissingNode` as in the main comment component: https://github.com/orbiting/styleguide/blob/2adf2bb22abcc83287873cb50ad0cdcbfcce3bab/src/components/Comment/Comment.js#L40

uxengine: @lukasbuenger I'll take care of this.

PR #149, opened 2018-07-22
VideoPlayer convenience

patte:
conveniences - a nice word for: fixes

- [x] play/pause with the spacebar and "k”
  only the VideoPlayer who has the focus toggles

- [x] fix mute icon sync between nativ and react components

### DIFF
--- a/src/components/VideoPlayer/Player.js
+++ b/src/components/VideoPlayer/Player.js
@@ -201,8 +201,6 @@ class VideoPlayer extends Component {
       }
     }
     this.scrubStart = event => {
-      this.captureFocus()
-
       this.scrubbing = true
       if (event.type === 'mousedown') {
         const up = e => {
@@ -397,7 +395,6 @@ class VideoPlayer extends Component {
                   globalState.instances.forEach(setter => {
                     setter(next)
                   })
-                  this.captureFocus()
                 }}
               >
                 <Subtitles off={!subtitles} />
@@ -410,7 +407,6 @@ class VideoPlayer extends Component {
                 e.preventDefault()
                 e.stopPropagation()
                 this.setMuted(!muted)
-                this.captureFocus()
               }}
             >
               <Volume off={muted} />
--- a/src/components/VideoPlayer/Player.js
+++ b/src/components/VideoPlayer/Player.js
@@ -271,7 +271,7 @@ class VideoPlayer extends Component {
       this.toggle()
     }
   }
-  captureFocus(event) {
+  captureFocus() {
     this.video.focus()
   }
   componentDidMount() {
@@ -336,7 +336,7 @@ class VideoPlayer extends Component {
 
     return (
       <div {...merge(styles.wrapper, breakoutStyles[size])}
-        onClick={this.captureFocus.bind(this)}
+        onClick={() => {this.captureFocus()}}
       >
         <video
           {...(isFullscreen ? styles.videoFullscreen : styles.video)}
@@ -352,7 +352,7 @@ class VideoPlayer extends Component {
           crossOrigin="anonymous"
           poster={src.thumbnail}
           tabIndex="0"
-          onKeyDown={this.handleKeyDown.bind(this)}
+          onKeyDown={(event) => {this.handleKeyDown(event)}}
         >
           <source src={src.hls} type="application/x-mpegURL" />
           <source src={src.mp4} type="video/mp4" />
--- a/src/components/VideoPlayer/Player.js
+++ b/src/components/VideoPlayer/Player.js
@@ -33,6 +33,9 @@ const styles = {
     },
     '::-webkit-media-controls-start-playback-button': {
       display: 'none !important'
+    },
+    ':focus': {
+      outline: 'none'
     }
   }),
   videoFullscreen: css({
--- a/src/components/VideoPlayer/Player.js
+++ b/src/components/VideoPlayer/Player.js
@@ -167,6 +167,11 @@ class VideoPlayer extends Component {
         loading: false
       }))
     }
+    this.onVolumeChange = () => {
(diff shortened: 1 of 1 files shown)


patte:
- can someone please check if I didn't dig a memory leak.
- does this block interaction in the publikator? @lukasbuenger

lukasbuenger:
From looking at the code, nothing suggests any broken behavior in the Publikator.

src/components/VideoPlayer/Player.js
```diff
@@ -167,6 +167,11 @@ class VideoPlayer extends Component {
         loading: false
       }))
     }
+    this.onVolumeChange = () => {
+      if(globalState.muted !== this.video.muted) {
```
uxengine: Let's consistently have a space after the if:  `if (globalState`

src/components/VideoPlayer/Player.js
```diff
@@ -369,13 +409,8 @@ class VideoPlayer extends Component {
...
                 e.stopPropagation()
-                const next = {
-                  muted: !muted
-                }
-                globalState.muted = next.muted
-                globalState.instances.forEach(setter => {
-                  setter(next)
-                })
+                this.setMuted(!muted)
+                this.captureFocus()
```
uxengine: I'm not fully convinced from an accessibility/screenreader perspective whether it's good to move focus back on the video after interacting with player controls. I'd tend to leave that decision to the user.

src/components/VideoPlayer/Player.js
```diff
@@ -302,7 +337,9 @@ class VideoPlayer extends Component {
     const { playing, progress, muted, subtitles, loading, fullscreen, isFullscreen } = this.state
 
     return (
-      <div {...merge(styles.wrapper, breakoutStyles[size])}>
+      <div {...merge(styles.wrapper, breakoutStyles[size])}
+        onClick={this.captureFocus.bind(this)}
+      >
         <video
           {...(isFullscreen ? styles.videoFullscreen : styles.video)}
```
uxengine: The focus triggers an outline on most browsers, e.g. a blue box shadow on Chrome:
screen shot 2018-07-23 at 10 19 41

You can suppress that by applying
```
':focus': {
  outline: 'none'
}
```

src/components/VideoPlayer/Player.js
```diff
@@ -302,7 +337,9 @@ class VideoPlayer extends Component {
     const { playing, progress, muted, subtitles, loading, fullscreen, isFullscreen } = this.state
 
     return (
-      <div {...merge(styles.wrapper, breakoutStyles[size])}>
+      <div {...merge(styles.wrapper, breakoutStyles[size])}
+        onClick={this.captureFocus.bind(this)}
```
uxengine: We tend to avoid bind() when possible. If this doesn't work
`onClick={this.captureFocus}`
could you try
`onClick={() =>{ this.captureFocus()}}`

src/components/VideoPlayer/Player.js
```diff
@@ -369,13 +409,8 @@ class VideoPlayer extends Component {
               onClick={e => {
                 e.preventDefault()
                 e.stopPropagation()
-                const next = {
-                  muted: !muted
-                }
-                globalState.muted = next.muted
-                globalState.instances.forEach(setter => {
```
uxengine: I'm sure Thomas had a reason when doing this, probably multiple players on the same page. Right now it looks like muting one player also changes the muted state of other players on the page. We should probably keep that per player as it is now.

src/components/VideoPlayer/Player.js
patte: sorry for that one, standard usually does that for me in precommit...

src/components/VideoPlayer/Player.js
patte: It's still here: https://github.com/orbiting/styleguide/pull/149/files#diff-02fd9ba94dd5c50faf6fb95bafcf4ec6R257

src/components/VideoPlayer/Player.js
patte: Thanks for this one. I removed captureFocus from all interactions.

src/components/VideoPlayer/Player.js
patte: I made the requested change. But can you elaborate or point me to a document on why bind should be avoided?

lukasbuenger: Binding is ok, when not happening in an update cycle. Binding in the render method will create and pass a new function every time an update is happening. So if you want to bind, do it in the contructor:
```
constructor(props) {
   super(props)
   this.captureFocus = this.captureFocus.bind(this)
   // or
   this.captureFocus = () => { /* your business logic*/ }
}
```

lukasbuenger: This however is also true for any anonymous arrow function, so the argument is somewhat tainted anyway, because that we do all the time and all over the place.

patte: Alright, I defined the function in the constructor.

src/components/VideoPlayer/Player.js
```diff
@@ -263,12 +293,14 @@ class VideoPlayer extends Component {
...
       return
     }
+
     this.video.addEventListener('play', this.onPlay)
     this.video.addEventListener('pause', this.onPause)
     this.video.addEventListener('loadstart', this.onLoadStart)
     this.video.addEventListener('canplay', this.onCanPlay)
     this.video.addEventListener('canplaythrough', this.onCanPlay)
     this.video.addEventListener('loadedmetadata', this.onLoadedMetaData)
+    this.video.addEventListener('volumechange', this.onVolumeChange)
```
tpreusse: missing removeEventListener in `componentWillUnmount`

PR #152, opened 2018-07-30
fix(Gallery): preload gallery images on activation

garamond:


src/components/Gallery/Gallery.js
```diff
@@ -175,9 +175,11 @@ class Gallery extends Component {
   render() {
     const { index, exitLeft, exitRight, closing, focus } = this.state
     const { onClose, items } = this.props
-    const { src, caption, credit, alt } = items[index]
+    const { src, caption, credit } = items[index]
```
tpreusse: `const currentItem = items[index]` would lead to more readable code down below.

src/components/Gallery/Gallery.js
```diff
@@ -175,9 +175,11 @@ class Gallery extends Component {
...
     const total = this.props.items.length
-    const srcs = FigureImage.utils.getResizedSrcs(src, window.innerWidth)
+    const reorderedItems = items.slice(index).concat(items.slice(0,index))
+    const preloadItems = reorderedItems.slice(0,3).concat(reorderedItems.slice(-2))
```
tpreusse: maybe something like this would be more explicit and easier to understand:
```js
[
  items[index], // aka currentItem if you sign that above
  items.slice(index + 1 , index + 3), // next 2
  items.slice(Math.max(0, index - 2), index) // prev 2
]
```

src/components/Gallery/Gallery.js
```diff
@@ -211,7 +213,20 @@ class Gallery extends Component {
...
+                {
+                  preloadItems.map((item, i) => {
+                    const srcs = FigureImage.utils.getResizedSrcs(item.src, window.innerWidth)
+                    return (
+                      <img 
+                        key={item.src} 
+                        alt={item.alt} 
+                        {...styles.mediaItemImage} 
+                        {...srcs} 
+                        style={{ display: src===item.src ? 'block' : 'none' }}
```
tpreusse: `src === item.src`

src/components/Gallery/Gallery.js
garamond: I agree.

PR #156, opened 2018-08-06
Feuilleton front

uxengine:
Main changes:
- optional `subject` element in front of lead
- a `framed` variant of FrontImage (container margin and switch to black headline in stacked mode)
- a `framed` variant of FrontSplit (basically the existing non-portrait teaser with adjusted margins)
- FrontTile updates for three tiles in a row: smaller headline and subject/lead font sizes

FYI, Luki's request for this was having the `subject` (better names welcome) as a block-level element on the markdown level in Publikator. That's why subject + lead have turned into inline elements. Open for better suggestions.

### The `subject` element in a nutshell:
- Will use H3 on the markdown level (placed between title and lead)
- It's optional and shall be used in Feuilleton teasers only
- All front teaser leads will use 22px on desktop now
- Subject and lead are inline elements now
- Design's gray by default
- Will adapt to teaser color if set and !== black (uncommon use case since feuilleton will typically be black on white) 

#### Desktop
screen shot 2018-08-08 at 14 40 53

screen shot 2018-08-13 at 10 51 30

screen shot 2018-08-13 at 10 59 46

#### Mobile
screen shot 2018-08-07 at 17 15 40

screen shot 2018-08-13 at 10 52 30

screen shot 2018-08-13 at 11 00 37

#### Styleguide
screen shot 2018-08-13 at 11 09 41

### DIFF
--- a/src/components/TeaserFront/Text.js
+++ b/src/components/TeaserFront/Text.js
@@ -134,6 +134,7 @@ Text.propTypes = {
     'bottom'
   ]),
   color: PropTypes.string,
+  compactColor: PropTypes.string,
   maxWidth: PropTypes.string,
   margin: PropTypes.string
 }
--- a/src/components/TeaserFront/Text.js
+++ b/src/components/TeaserFront/Text.js
@@ -92,7 +92,7 @@ const Text = ({
   position,
   center,
   color,
-  compactColor,
+  collapsedColor,
   maxWidth,
   margin
 }) => {
@@ -100,8 +100,8 @@ const Text = ({
   const rootStyles = position ? styles.rootPosition : {}
   const middleStyles = position === 'middle' ? styles.rootMiddle : {}
 
-  const colorStyle = compactColor && css({
-    color: compactColor,
+  const colorStyle = collapsedColor && css({
+    color: collapsedColor,
     [tUp]: {
       color
     }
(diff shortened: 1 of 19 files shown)


uxengine:
@tpreusse Ready to review now!

src/components/Typography/Editorial.js
```diff
@@ -49,27 +49,66 @@ export const Subhead = ({ children, attributes, ...props }) => (
   </h2>
 )
 
-const lead = css({
+const leadStyle = {
   ...styles.serifRegular19,
+  lineHeight: '27px',
   margin: '0 0 10px 0',
   [mUp]: {
```
tpreusse: mv `[mUp]` to `const lead`. Will be completely overwritten in `leadSmall`. If you want margin there too add it there explicitly.

uxengine:
@tpreusse All comments addressed (except mediaQueries refactoring). Most importantly, lead and subject styles are now completely moved to the `TeaserFront` namespace and the article lead remains untouched. Confirmed with design that all regular desktop leads should use `23px font-size / 30px line-height`.

src/components/TeaserFront/Text.js
```diff
@@ -92,18 +92,28 @@ const Text = ({
...
+  })
+
   return (
     <div {...rootStyles} {...middleStyles}>
       <div
         {...attributes}
+        {...colorStyle}
         {...css(styles.positioned, position ? styles[position] : {})}
-        style={{ color, textAlign, maxWidth, margin }}
+        style={{ color: !collapsedColor && color, textAlign, maxWidth, margin }}
```
tpreusse: This may result in `color:;text-align:` (`style={{color: false, textAlign: ''}}`). Use `undefined` instead of `false` and `''` (`style={{color: undefined, textAlign: undefined}}` results in no style attribute at all).

PR #158, opened 2018-08-16
feat(VideoPlayer): isCinemagraph flag

uxengine:
This hit us again when introducing the autoPlay flag on Publikator, which doesn't autoplay the video on iOS only. I'd vote for adding the cross-platform magic here.


tpreusse:
As discussed in person: Let move to a `cinemagraph` property in Publikator and then set all the required props for it in the article schema when mapping to `VideoPlayer` .

uxengine:
See https://r-styleguide-pr-158.herokuapp.com/videoplayer#iscinemagraph

PR #159, opened 2018-08-16
feat(Logo): expose ratio

tpreusse:
Useful to support different logos on `.love`:

screen shot 2018-08-16 at 18 00 01


uxengine:
L.O.V.E.

PR #163, opened 2018-08-21
feat(Article): support Subject in article titleblock

uxengine:
#### Desktop
screen shot 2018-08-22 at 10 10 22

#### Mobile
screen shot 2018-08-22 at 09 50 33


uxengine:
@tpreusse Just bumping this up in your inbox :)

PR #172, opened 2018-09-11
fix(components/Form/Autocomplete): show a focused label if !!value

lukasbuenger:
image

... instead of ...

image

### DIFF
--- a/src/components/Form/Autocomplete.js
+++ b/src/components/Form/Autocomplete.js
@@ -45,7 +45,7 @@ const Autocomplete = ({
             <Inner isOpen={isOpen}>
               <Field
                 label={label}
-                value={isOpen ? filter : value.text}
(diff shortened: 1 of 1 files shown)


src/components/Form/Autocomplete.js
```diff
@@ -44,9 +44,8 @@ const Autocomplete = ({
           <div {...styles.root}>
             <Inner isOpen={isOpen}>
               <Field
-                isFocused={isOpen}
                 label={label}
-                value={filter}
+                value={isOpen ? filter : value.text}
```
tpreusse:
```
TypeError: Cannot read property 'text' of null
```

Happens on the initial load of the styleguide doc page.

lukasbuenger: Phew. Will fix asap.

PR #181, opened 2018-10-15
feat(Comment): misc improvements (collapsing, upvotes/downvotes, share)

uxengine:
Related to
https://github.com/orbiting/republik-frontend/pull/200
https://github.com/orbiting/backends/pull/118

#### Three features:
- add optional "collapsable" UI 
- explicit upvotes/downvotes instead of single score
- share icon in actions bar

#### Collapsing logic in a nutshell:
- A discussion can be flagged as `collapsable` in Publikator
- If a discussion is collapsable, the comment component measures whether its body is significantly higher than the max height and therefore `shouldCollapse`, and reports that back to the HOC row
- the row passes `collapsed` prop down to `Comment` and `CommentActions` (note that `true/false` triggers the respective UI while `undefined` triggers nothing)

#### Styleguide deploy
https://r-styleguide-pr-181.herokuapp.com/components/comment#ltcommentactions-gt

#### in republik-frontend
screen shot 2018-10-17 at 13 54 01
screen shot 2018-10-17 at 13 54 30
screen shot 2018-10-17 at 13 55 06

#### Ellipsizing with all icons present on small screen
screen shot 2018-10-17 at 13 56 27

### DIFF
--- a/src/components/Comment/Comment.js
+++ b/src/components/Comment/Comment.js
@@ -1,10 +1,11 @@
-import React from 'react'
+import React, { Component } from 'react'
 import { css } from 'glamor'
 import { renderMdast } from 'mdast-react-render'
 
 import { Label } from '../Typography'
 
 import colors from '../../theme/colors'
+import { mBreakPoint, mUp } from '../../theme/mediaQueries'
 
 import createCommentSchema from '../../templates/Comment'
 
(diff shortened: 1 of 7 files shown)


uxengine:
@tpreusse 

Thanks for your suggestions. We had discussed and decided not to use "Mehr" initially because in the context of the other icons on the left, it sounded like it would reveal more icons and not more of the comment. However, it could work when combined with an expand/collapse icon. 

So here's a suggestion:
- The expand/collapse buttons now have an icon (i,e. the former vote up/down icons). On mobile, only the icon remains.
- The upvote/down buttons now use `MdArrowDropUp/Down`

screen shot 2018-10-22 at 16 49 42
screen shot 2018-10-22 at 16 50 20
screen shot 2018-10-22 at 16 51 28

PR #182, opened 2018-10-18
feat(FigureCover): add breakout size

uxengine:
screen shot 2018-10-18 at 18 18 52
screen shot 2018-10-18 at 18 20 01


src/components/Figure/index.js
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { css, merge } from 'glamor'
 import { mUp } from '../../theme/mediaQueries'
-import {
+import Center, {
```
tpreusse: Is `Center` an unused import?

PR #186, opened 2018-11-01
fix(CommentActions): hairline only if comment is collapsed

uxengine:
Minor UI fix which makes collapsed and expanded state better distinguishable without reading the label. (Think: There's a line, so it can be expanded.) Suggested by users.

#### Before
screen shot 2018-11-01 at 15 58 00

#### After
screen shot 2018-11-01 at 15 57 08

### DIFF
--- a/src/components/Comment/CommentActions.js
+++ b/src/components/Comment/CommentActions.js
@@ -91,7 +91,7 @@ const styles = {
     lineHeight: `${config.left}px`
   }),
   collapsed: css({
-    borderTop: `1px solid ${colors.divider}`,
+    borderTop: `1px solid ${colors.primary}`,
     paddingTop: '6px'
   }),
   centerButton: css({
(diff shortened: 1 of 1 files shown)


src/components/Comment/CommentActions.js
```diff
@@ -122,7 +122,7 @@ export const CommentActions = ({
   const collapsable = collapsed !== undefined
   const collapseLabel = t(`styleguide/CommentActions/${ collapsed ? 'expand' : 'collapse'}`)
   return (
-    <div {...styles.root} {...(collapsable && !highlighted ? styles.collapsable : undefined)}>
+    <div {...styles.root} {...(collapsable && collapsed && !highlighted ? styles.collapsed : undefined)}>
```
garamond: I think `{...(collapsable && collapsed && !highlighted && styles.collapsed)}`would work as well

uxengine: I thought spreading `...(false)` is nor supported, while `...(undefined)` is?

PR #206, opened 2018-12-20
feat(Article): add attributes for tracking reading/watching/listening progress

uxengine:
Required for tracking reading progress. Adds a data attribute to 
- most* first-level article elements (i.e. children of `Center`)
- edge-to-edge images (root-level)
- but not: elements in cover and title block.

_*) a few less important article element types haven't been tagged yet, but it's always possible to do later because the child index is based on mdast node index, and not the sequence of tagging._

`data-pos="{parentIndex}-{childIndex}"` or `data-pos="{nodeIndex}"` (on root-level)
For storing a specific element ID which represents reading progress. It's based on the parent's and child's mdast node index. Parent index is needed because edge-to-edge images interrupt the normal article element structure and inject a `Center` on the article root-level.

The ID generation pattern is still up for discussion.

Example *with* edge-to-edge image:
screenshot 2019-01-23 at 18 51 53

Example *without* edge-to-edge image:
screenshot 2019-01-23 at 18 51 22

### Audio 

image

### Video

image


src/components/AudioPlayer/Player.js
```diff
@@ -251,8 +257,16 @@ class AudioPlayer extends Component {
...
       }))
     }
+    this.setTime = (time = 0) => {
+      if (this.audio.currentTime !== time) {
+        this.audio.currentTime = time
+        this.updateProgress()
+      }
+    }
     this.onCanPlay = () => {
+      !this.state.initialized && this.setTime(this.props.startSeconds)
```
tpreusse: also add this to `onLoadedMetaData`, iOS apparently does not fire `onCanPlay`.

src/components/AudioPlayer/Player.js
```diff
@@ -553,7 +575,8 @@ AudioPlayer.propTypes = {
   download: PropTypes.bool,
   scrubberPosition: PropTypes.oneOf(['top', 'bottom']),
   timePosition: PropTypes.oneOf(['left', 'right']),
-  controlsPadding: PropTypes.number
+  controlsPadding: PropTypes.number,
+  time: PropTypes.number,
```
tpreusse: probably outdated, `startSeconds`?

src/components/AudioPlayer/Player.js
```diff
@@ -197,7 +202,8 @@ class AudioPlayer extends Component {
       progress: 0,
       loading: false,
       buffered: null,
-      sourceError: false
+      sourceError: false,
+      initialized: false,
```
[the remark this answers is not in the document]
uxengine: I've removed that now in favor of `contextStartSeconds`.

src/components/AudioPlayer/Player.js
uxengine: I've created `initStartTime` now which is called both from `onCanPlay`, `onLoadedMetaData`, `componentDidMount` and `componentDidUpdate`. The first two event handlers may not even fire in some cases (e.g. when opening a new tab on desktop with the same article).

Issue #174, opened 2018-09-21
feat(components/Form/Field): Remove isFocused


lukasbuenger:
Remove property `isFocused` from `Field` component. 

Discussion and context: https://github.com/orbiting/styleguide/pull/172

PR #211, opened 2019-01-16
Field Cleanup

tpreusse:
- remove no longer used `isFocused` prop https://github.com/orbiting/styleguide/issues/174
- fix label overlap with `0` value
- clean up docs a bit

screenshot 2019-01-16 at 11 36 55

### DIFF
--- a/src/components/Form/Field.js
+++ b/src/components/Form/Field.js
@@ -153,7 +153,6 @@ class Field extends Component {
       renderInput,
       onInc,
       onDec,
-      isFocused: isFocusedFromProps,
       icon
     } = this.props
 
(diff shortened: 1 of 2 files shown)


src/components/Form/Field.js
```diff
@@ -182,7 +180,8 @@ class Field extends Component {
     const hasIncrease = !!onInc
     const hasDecrease = !!onDec
     const hasError = !!error
-    const labelStyle = (isFocused || value || hasError)
+    const valueIsPresent = value !== undefined && value.length !== 0
```
uxengine: This would explode if any consuming code would set the `value` prop to `null` for some reason. We probably don't do that, but might be good to double-check.

src/components/Form/docs.md
```diff
@@ -4,38 +4,34 @@
...
-      size={30}
-      onClick={() => {
-        console.log('search')
-      }}
-    />
-  } />
+  icon={<SearchIcon size={30} />} />
 ```
 
+Normally `value` should be a string. Even if you pass in numbers, you'll recieve strings on change by default.
```
uxengine: Typo: recieve -> receive

PR #214, opened 2019-02-12
dynamic-component: externalComponents

patte:
This allows to provide `externalComponents` to the `Article` schema upon init. If `DynamicComponent` encounters a `src` it can find in `externalComponents` the externalComponent is used instead of trying to require a url. With this solution for example `republik-frontend` components can be rendered in an article.

See: https://github.com/orbiting/republik-frontend/pull/256/files#diff-28ad1650efde907abc8651aa47fe035a

Example md:
```
## Geld
<section><h6>DYNAMIC_COMPONENT</h6>

\```
{
  "autoHtml": true,
  "props": {
    "discussionId": "86da7294-f345-4cb4-a269-0d9e2ef06f65"
  },
  "src": "QuestionSource"
}
\```
<hr /></section>
```


patte:
Closed, as providing the external component [to `dynamicComponentRequire`](https://github.com/orbiting/republik-frontend/blob/master/components/Article/Page.js#L84) which is consumed in styleguide [here](https://github.com/orbiting/styleguide/blob/master/src/components/DynamicComponent/index.js#L39) was all I needed to make it work.

PR #219, opened 2019-02-26
fix(MediaPlayers): load user progress from promise and remove unused startSeconds prop

uxengine:
Removes the unused `startSeconds` prop and loads startSeconds via promise provided by frontend context.

### DIFF
--- a/src/components/VideoPlayer/Player.js
+++ b/src/components/VideoPlayer/Player.js
@@ -318,25 +318,18 @@ class VideoPlayer extends Component {
       this._textTrackMode = subtitles
     }
   }
-  getStartTime() {
-    return new Promise((resolve, reject) => {
-      if (this.props.mediaId && this.context.getMediaProgress) {
-        this.context.getMediaProgress(this.props.mediaId)
-          .then((startSeconds) => {
-            if (startSeconds) {
-              this.setState(() => ({ startSeconds }))
-              !!startSeconds && this.setTime(startSeconds)
-            }
-            resolve()
-          }
-        ).catch(() => {
-          resolve()
-        })
-      } else {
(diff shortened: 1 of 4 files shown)


src/components/VideoPlayer/Player.js
```diff
@@ -319,22 +318,25 @@ class VideoPlayer extends Component {
...
-      return
-    }
-    let startSeconds
-    if (this.props.mediaId && this.context.getMediaProgress) {
-      startSeconds = this.context.getMediaProgress(this.props.mediaId)
-      if (startSeconds) {
-        this.setState(() => ({
-          contextStartSeconds: startSeconds
-        }))
+  getStartTime() {
```
tpreusse: `initStartTime` IMO was better since this gets and sets the start time.

tpreusse: or make it resolve to `startSeconds` or `undefined` and leave the get name

src/components/VideoPlayer/Player.js
```diff
@@ -319,22 +318,25 @@ class VideoPlayer extends Component {
...
+    return new Promise((resolve, reject) => {
```
tpreusse: Simpler version:
```js
if (this.props.mediaId && this.context.getMediaProgress) {
  return this.context.getMediaProgress(this.props.mediaId).then((startSeconds) => {
    if (startSeconds) {
      this.setState(() => ({ startSeconds }))
      !!startSeconds && this.setTime(startSeconds)
    }
  }).catch(() => {
    return undefined // ignore errors
  })
}
return Promise.resolve()
```

uxengine:
Thanks! Comments addressed.

PR #234, opened 2019-05-09
Fix TeaserFrontTileRow layout

garamond:
- Derive layout from number of children (remove `columns` prop)
- Add `expand` prop for teasers on front
- Adjust breakpoints

| Desktop | Tablet| Mobile |
| -- | -- | -- |
|  |  |  |


garamond:
Superseeded by #237

PR #239, opened 2019-05-23
Dialog 4.0

wereHamster:
I think it's best if you start looking through the Discussion related pages in the catalog, and then from there dive into the code.

* Upgraded catalog and react-scripts (the `@catalog/core` package no longer depends on webpack)
* Upgraded react to 16.8 due to use of React Hooks. Currently use two hooks: `useSize()` and `useMediaQuery()`. The `useSize()` hook depends on ResizeObserver though.
* DiscussionContext (React Context) provides almost all discussion-wide static data and mutation callbacks to the comment tree components. This greatly simplifies the components as they take fewer direct props and don't have to pass through so much. The `t` prop could also be packaged up in the DiscussionContext, but I decided not to, it's still passed explicitly where required.
* Many components now take a single `comment` prop (which must have the same shape as the GraphQL `Comment` type) instead of taking individual props. This also decreases the number of props that have to be passed through. For documentation purposes, we have a few pre-made comments in different states (`comment1`… `comment9`).

Because the repo now uses a newer react-scripts version, which comes with a newer version of babel, you can now use `<>` instead of `<Fragment>`.


src/components/Discussion/Internal/Comment/Body.js
```diff
@@ -0,0 +1,173 @@
...
+   * Measuring the comment body size (height), so we can deterimine whether to collapse
+   * the comment body.
+   *
+   * bodyVisibility:
+   *   - 'indeterminate': We don't know yet whether to collapse the body or not.
+   *   - 'full': The body is collapsable but we're showing the full body.
+   *   - 'preview': The body is collapsed.
+   */
+  const [bodyVisibility, setBodyVisibility] = React.useState('indeterminate')
+  const [bodyRef, bodySize] = useSize()
```
tpreusse: Use an `useEffect(() => {}, [content])` that simply measures the body once after rendering different content instead. I don't see any need for a resize observer.

This of course can be abstracted as `useMeasureOnce` or something like that. Or even call it `useSize`, if you want, but with content change dependency ;)

src/components/Discussion/Internal/Comment/Body.js
```diff
@@ -0,0 +1,173 @@
...
+    [mUp]: {
+      maxHeight: `${COLLAPSED_HEIGHT.desktop}px`
+    }
+  }),
+  context: css({
+    marginBottom: 10
+  }),
+  collapeToggleContainer: css({
+    borderTop: `1px solid ${colors.divider}`,
+    marginTop: 6
```
tpreusse:
```suggestion
    marginTop: 0
```
Screenshot 2019-05-28 at 02 10 46

src/components/Discussion/Internal/Comment/Body.js
wereHamster: I created a new hook: `useBoundingClientRect()`. The old hook is still available, in case you want to use it at some point in the future.

src/components/Discussion/Internal/Comment/Header.js
```diff
@@ -0,0 +1,257 @@
...
+    }
+  }),
+  action
+}
+
+const dateTimeFormat = timeFormat('%d. %B %Y %H:%M')
+const titleDate = string => dateTimeFormat(new Date(string))
+
+export const Header = ({ t, comment, isExpanded, onToggle }) => {
+  const { clock, links } = React.useContext(DiscussionContext)
```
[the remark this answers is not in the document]
wereHamster: I removed the default value from the DiscussionContext. Components which require the context will now likely crash hard if you don't wrap them in the provider.

PR #248, opened 2019-07-16
Data points and tooltips

garamond:
image

Add data with tooltips to points on a map. 

New props:

 * `pointLabel`: CSV column to read the tooltip header from
 * `pointTooltips`: Ordered list of CSV columns to display in tooltip body

[Documentation](https://r-styleguide-pr-248.herokuapp.com/charts/maps#map-with-data-points-and-tooltips)


src/components/Chart/Maps.js
```diff
@@ -189,6 +221,56 @@ export class GenericMap extends Component {
...
+    const [ x, y ] = projectPoint([hoverPoint.datum.lon, hoverPoint.datum.lat])
+
+    const value = isNaN(hoverPoint.datum.value) 
+      ? new String(hoverPoint.datum.value).trim()
+      : numberFormat(hoverPoint.datum.value)
+
+    const body = pointTooltips.map(t => {
+      const val = hoverPoint.datum[t]
+      if (val) {
+        return (<>{`${t}: ${ isNaN(val) ? val : numberFormat(+val) }`}<br/></>)
```
tpreusse: I like the neatness of using the column name as a label / unit.

However the current implementation limits everything to this style:
> **New York**
> 1784 km<sup>2</sup>
> Einwohner: 862'300

But normally we use the unit as a suffix:
> **New York**
> 1784 km<sup>2</sup>
> 862'300 Einwohner

Also we might have different number formatting requirements for the different values. E.g.
> **New York Area**
> 17,1 Tonnen CO<sub>2</sub> pro Einwohner
> 13,6 Mio. Einwohner

First line would be `.1f` and second `.3s`. I don't think it worth abstracting all those problems and would just expect pre formatted details lines, in this case `13,6 Mio. Einwohner` – no type detecting, casting or `numberFormat` here. But do pipe those detail lines through `subsup` to support fancy stuff like CO<sub>2eq</sub>.

src/components/Chart/Maps.js
```diff
@@ -39,30 +52,52 @@ const symbolShapes = {
...
 const shapes = Object.keys(symbolShapes).concat('marker')
 
-const Points = ({data, colorScale, colorAccessor, project, shape, sizes, domain}) => {
+const Points = ({data, colorScale, colorAccessor, project, shape, sizeRangeMax, hoverPoint, setHoverPoint}) => {
+
   const marker = shape === 'marker'
   let symbolPath
   if (!marker) {
-    const size = scaleOrdinal().domain(domain).range(sizes)
+    const size = scaleLinear().domain(extent(data.map(d => d.value))).range([0, sizeRangeMax])
```
tpreusse: Add zero to the domain:

```suggestion
    const size = scaleLinear().domain([0, max(data.map(d => d.value))]).range([0, sizeRangeMax])
```

Makes Mumbai with 1t visible

src/components/Chart/Maps.layout.js
```diff
@@ -74,35 +75,8 @@ export default (props, geoJson) => {
       .domain(domain)
       .range(colorRange)
   } else {
-    const dataValues = data.map(d => d.value)
```
tpreusse: Accidental delete?

Screenshot 2019-07-30 at 12 05 51

src/components/Chart/Maps.js
```diff
@@ -189,6 +227,56 @@ export class GenericMap extends Component {
...
+    const [ x, y ] = projectPoint([hoverPoint.datum.lon, hoverPoint.datum.lat])
+
+    const value = isNaN(hoverPoint.datum.value) 
+      ? String(hoverPoint.datum.value).trim()
+      : numberFormat(hoverPoint.datum.value)
+
+    const body = pointDetails.map(t => {
+      const val = hoverPoint.datum[t]
+      if (val) {
+        return (<>{`${t}: ${ isNaN(val) ? val : numberFormat(+val) }`}<br/></>)
```
tpreusse:
```suggestion
        return (<>{subsup(val)}<br/></>)
```

We should not force the `Column: Value` format and also no number formatting. Just assume that the details are strings and if they contain numbers they are correctly pre formatted.

src/components/Chart/Maps.layout.js
```diff
@@ -52,7 +53,7 @@ export default (props, geoJson) => {
   const numberFormat = getFormat(props.numberFormat, t)
   let domain
   let colorScale
-  let colorAccessor = d => d.value
+  let colorAccessor = d => d[color] || 0
```
tpreusse: `choropleth` are still broken, most likely because of this.

```suggestion
  let colorAccessor = choropleth ? d => d.value : d => d[color] || 0
```

tpreusse:
Please test the whole styleguide page: https://r-styleguide-pr-248.herokuapp.com/charts/maps
and make sure stuff still works.

garamond:
Works again, sorry!

src/components/Chart/Maps.js
```diff
@@ -189,6 +226,70 @@ export class GenericMap extends Component {
...
+    if (showTooltip) {
+      return (
+        <ContextBox
+          orientation="top"
+          x={x}
+          y={y}
+          contextWidth={width}
+        >
+          <ContextBoxValue label={hoverPoint.datum[pointLabel]}>
+            <div {...styles.tooltip}>
```
tpreusse: `<ContextBox>` already sets fontSize 14 and `WebkitUserSelect` is not needed here right? IMO rm.

PR #251, opened 2019-08-16
Move gallery to front-end

garamond:


src/components/Figure/Image.js
```diff
@@ -26,29 +53,44 @@ class Image extends Component {
...
+    if (src && gallerySize > 0) {
+      wrappedImage =
+        <>
+          {wrappedImage}
+          <GalleryButton gallerySize={gallerySize} onClick={handleClick} />
+        </>
+    }
+    return (
+      <div {...styles.imageContainer}>
+        {wrappedImage}
```
tpreusse:
```suggestion
        {wrappedImage}
        {gallerySize > 0 && <GalleryButton gallerySize={gallerySize} onClick={handleClick} />}
```
and rm the `if (src && gallerySize > 0)` above

src/components/Figure/Image.js
```diff
@@ -26,29 +53,44 @@ class Image extends Component {
       maxWidth,
       size: sizeProp,
       aboveTheFold,
-      enableGallery = true
+      enableGallery = true,
+      gallerySize
     } = this.props
 
     const onClick = (enableGallery && this.context.toggleGallery) ? this.context.toggleGallery : () => {}
```
tpreusse:
```suggestion
     const isClickable = (enableGallery && this.context.toggleGallery)
     const onClick = isClickable
      ? () => this.context.toggleGallery(src)
      : () => {}
```
create a function to call `toggleGallery` with `src` here and then directly use `onClick` instead of an additional `handleClick` handler.

Also add a cursor `pointer` or `zoom-in` if `isClickable`.

PR #264, opened 2019-10-29
feat(Button): make primary button hover style pop more

annatraussnig:
- Replace the dark hover color with a more subtle shadow.
- Disable hover state color changes for disabled buttons

Screenshot 2019-10-31 at 15 52 17

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,14 @@ This is a living style guide. Subject to constant change.
 
 It documents the current state and provides implemented React components, published as an npm package.
 
+## About
+
+The development started on March 5th 2017 for the crowdfunding of [republik.ch](https://www.republik.ch/). It has since been continuously expanded and now provides a wide range of functionality from [colors](/colors) and [typography](/typography) definitions, to an [video player](/videoplayer), common [form elements](/forms), [discussion trees](/components/discussion/tree), a full suite of editorial elements—from teasers to infoboxes, template definitions targeting [web](/templates/article) and [email](/templates/editorialnewsletter) and a [charting system](/charts).
+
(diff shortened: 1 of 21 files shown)


tpreusse:
Re: `primaryBright` to white contrast is 2.1:
https://color.review/check/FFFFFF-47CC00
We should hit at least 3.

Unfortunately even our current `primary` to white is only 2.9:
https://color.review/check/FFFFFF-3CAD00

I'd recommend to do something like this:
```js
import { lab } from 'd3-color'
const labColor = lab(colors.primary)
```
https://github.com/orbiting/styleguide/blob/77239ac34ddc8a9ea48944cdae672a1e3bbc85af/src/components/TeaserFront/CreditLink.js#L15-L21

Leading to `#099000` with an 4.1 contrast to white:
https://color.review/check/FFFFFF-099000

annatraussnig:
> `primaryBright` to white contrast is 2.1:
> https://color.review/check/FFFFFF-47CC00
> We should hit at least 3.
> 
> Unfortunately even our current `primary` to white is only 2.9:
> https://color.review/check/FFFFFF-3CAD00
> 
> I'd recommend to do something like this:
> 
> ```js
> import { lab } from 'd3-color'
> const labColor = lab(colors.primary)
> ```
> 
> https://github.com/orbiting/styleguide/blob/77239ac34ddc8a9ea48944cdae672a1e3bbc85af/src/components/TeaserFront/CreditLink.js#L15-L21
> 
> Leading to `#099000` with an 4.1 contrast to white:
> https://color.review/check/FFFFFF-099000

Done

PR #269, opened 2019-11-07
feat(Buttons): forward props to component

annatraussnig:
Allows us to apply glamor-generated css styles to buttons


tpreusse:
The problem with this is that additional glamor styles may or may not conflict with existing styles.
I would prefer not to forward as the pattern is error prone. For margin cases using child selectors to target the button is preferred. They also have more precedence and should win over definitions in the button.

PR #274, opened 2019-12-11
Custom paynotes

annatraussnig:
Add `paynotes` variable to templates. We use this variable to decide whether we should render the custom paynotes form in publikator


src/templates/Dossier/index.js
```diff
@@ -11,6 +11,7 @@ const createSchema = ({
   dossierHref = '/dossier',
   customMetaFields = [],
   series = false,
+  paynotes = false,
```
tpreusse: IMO should be true. They already have the same paynotes as articles.

PR #276, opened 2020-01-08
Dark mode in photo series

annatraussnig:
- Add darkMode flag to various schemas (atm only enabled for articles)
- Add ColorContext and useColorContext hook
- Use color context instead of colors for styling of a subset of editorial elements and article container


src/components/Colors/useColorContext.js
```diff
@@ -0,0 +1,17 @@
+import React, { useContext, useState, useEffect } from 'react'
+import ColorContext from './ColorContext'
+import colors from '../../theme/colors'
+
+export const useColorContext = () => {
+  const colorContext = useContext(ColorContext)
+  const [colorScheme, setColorScheme] = useState(colors)
+
+  useEffect(() => {
```
annatraussnig: @tpreusse please have a look at this one in particular – without useEffect I ran into the issue that the ColorContext wouldn't take effect as it should and that parts of the page kept their initial white bg (would happen now and then, not consistently), but maybe there is a more concise solution to this

PR #280, opened 2020-01-20
Dark mode for related articles

annatraussnig:
- handle default color scheme for teaser/tiles (based on color context)

**Needs to be tested for teasers with user set colors**

Screenshot 2020-01-20 at 17 53 33

### DIFF
--- a/src/components/TeaserFront/CreditLink.js
+++ b/src/components/TeaserFront/CreditLink.js
@@ -59,6 +59,7 @@ const CreditLink = ({
 
 CreditLink.propTypes = {
   children: PropTypes.node.isRequired,
+  color: PropTypes.string,
   collapsedColor: PropTypes.string
 }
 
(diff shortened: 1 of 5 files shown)


src/components/TeaserFront/CreditLink.js
```diff
@@ -65,12 +59,7 @@ const CreditLink = ({
 
 CreditLink.propTypes = {
   children: PropTypes.node.isRequired,
-  color: PropTypes.string,
```
tpreusse: This prop is still supported right? You scared me by removing the prop type :)

src/components/TeaserFront/Tile.js
```diff
@@ -147,8 +153,6 @@ Tile.propTypes = {
   byline: PropTypes.string,
   alt: PropTypes.string,
   onClick: PropTypes.func,
-  color: PropTypes.string,
-  bgColor: PropTypes.string,
```
tpreusse: same here, props still support – we should keep the prop types.

src/components/TeaserFront/CreditLink.js
annatraussnig: YES, still supported but I think having the string type turns `undefined` to `"undefined"`, which caused problems down the line – I'll double check that this is the case. In case it is, do you know how to fix that? (I'm not excited to do comparisons which look like `color === 'undefined' ?`)

PR #283, opened 2020-02-13
Add useHeaderHeight hook

garamond:


src/lib/useHeaderHeight.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+
+const DEFAULT_CONFIG = [{ minWidth: 0, headerHeight: 0 }]
+
+const HeaderHeightContext = React.createContext({
+  value: 0,
```
tpreusse:
```suggestion
  value: DEFAULT_CONFIG[0].headerHeight,
```

src/lib/useHeaderHeight.js
```diff
@@ -0,0 +1,55 @@
...
+  rules: []
+})
+
+export const useHeaderHeight = () => {
+  const headerHeightContext = React.useContext(HeaderHeightContext)
+  return [headerHeightContext.value, headerHeightContext.rules]
+}
+
+export const HeaderHeightProvider = ({ children, config = DEFAULT_CONFIG }) => {
+  const [headerHeightValue, setHeaderHeightValue] = React.useState(0)
```
tpreusse:
```suggestion
  const [headerHeightValue, setHeaderHeightValue] = React.useState(config[0].headerHeight)
```

src/lib/useHeaderHeight.js
```diff
@@ -0,0 +1,55 @@
...
+        mediaQuery: `@media only screen and (min-width: ${minWidth}px)`,
+        headerHeight
+      })),
+    [config]
+  )
+
+  const handleResize = () => {
+    if (window) {
+      const nextHeaderHeightValue = config.reduce((acc, cur) => {
+        if (window.innerWidth >= cur.minWidth && cur.minWidth >= acc) {
```
tpreusse: What does `cur.minWidth >= acc` do? Why ensure that the current `minWidth` is greater than the last `headerHeight`?

IMO you can assume (and document) that config has to be sorted in the desired order (normally by `minWidth` asc) and just assume the last one can win.

PR #289, opened 2020-03-17
fix(EmailList): use generic P instead of Editorial.P for web lists

annatraussnig:


src/templates/EditorialNewsletter/schema.js
```diff
@@ -72,39 +72,41 @@ const createNewsletterSchema = ({
...
-        matchMdast: matchType('strong'),
-        component: ({ attributes, children }) => (
-          <strong {...attributes}>{children}</strong>
-        ),
-        editorModule: 'mark',
-        editorOptions: {
-          type: 'STRONG',
-          mdastType: 'strong'
-        }
+  const paragraph = customComponent => {
```
tpreusse:
```suggestion
  const createParagraphRule = customComponent => {
```

and then just call it twice:
```js
const paragraph = createParagraphRule()
const listParagraph = createParagraphRule(ListP)
```

src/templates/EditorialNewsletter/schema.js
```diff
@@ -72,41 +72,46 @@ const createNewsletterSchema = ({
...
-          type: 'STRONG',
-          mdastType: 'strong'
-        }
+  const createParagraphRule = customComponent => {
+    return {
+      matchMdast: matchParagraph,
+      component: customComponent || Paragraph,
+      editorModule: 'paragraph',
+      editorOptions: {
+        formatButtonText: 'Paragraph'
```
tpreusse:
```suggestion
        formatButtonText: 'Paragraph',
        type: customComponent ? 'LISTP' : undefined
```

We need to set this custom type. Otherwise Publikator will not be able to render two different components (maybe change the abstraction to receive the type name via arg as well)

PR #295, opened 2020-04-21
fix(Article): remove progress data id from last elements of article

ovbm:


src/templates/Article/index.js
```diff
@@ -155,6 +155,20 @@ const getProgressId = (node, index, parent, { ancestors }) => {
     const rootNode = ancestors[ancestors.length - 1]
     const indexOfParent =
       rootNode && rootNode.children.length && rootNode.children.indexOf(parent)
+
+    // Do not add Progress ID for last element of last center component, if elements are infoboxes or articlecollections
+    if (
+      index === parent.children.length - 1 &&
+      indexOfParent === rootNode.children.length - 1
```
tpreusse, line 162:
```suggestion
      rootNode &&
      indexOfParent === rootNode.children.length - 1
```
Generally `rootNote` is always available but if I remember correctly in certain (test) environments it wasn't. Because it already coded defensively for `indexOfParent` I'd also add a defensive check before accessing `rootNode.children`.

ovbm: will do

PR #304, opened 2020-06-15
fix(audioplayer): loader position and client side Link

ovbm:
Screenshot 2020-06-15 at 11 44 44

### DIFF
--- a/src/components/AudioPlayer/Player.js
+++ b/src/components/AudioPlayer/Player.js
@@ -476,7 +476,7 @@ class AudioPlayer extends Component {
       maxWidth: `calc(100% - ${leftIconsWidth + rightIconsWidth + 20}px)`,
       left: timePosition === 'left' ? leftIconsWidth + 10 : 'auto',
       right: timePosition === 'right' ? rightIconsWidth + 10 : 'auto',
-      top: loading ? '0px' : fixed ? '-12px' : '1px'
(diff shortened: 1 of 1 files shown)


src/components/AudioPlayer/Player.js
```diff
@@ -623,16 +626,18 @@ class AudioPlayer extends Component {
             )}
             {!loading && title && fixed && (
               <div {...styles.time}>
-                <A style={{ color: colors.text }} href={sourcePath}>
+                <Link style={{ color: colors.text }} href={sourcePath} passHref>
                   {title}
```
tpreusse, line 630: You still need the A tag:
```suggestion
                <Link href={sourcePath} passHref>
                  <A style={{ color: colors.text }} href={sourcePath}>
                    {title}
                  </A>
```

See [Next.js docs](https://nextjs.org/docs/api-reference/next/link) for a bit of background. Although it's not that well explained. I think it's worth reading the render method of the next.js Link component:
https://github.com/vercel/next.js/blob/e4f0848c81df060fb557271d3035bbd6d1ba13ed/packages/next/client/link.tsx#L238-L307 (ignore the url / href formatting parts, we used [next-routes](https://www.npmjs.com/package/next-routes) and custom logic which preprocesses this)

ovbm: Ah! I was just at the cusp of going crazy XD

PR #314, opened 2020-07-27
Action Bar changes

ovbm:
- Introduces the IconButton component that replaces IconLink (incl docs)
- Refactors Callout to not be opinionated about the CalloutMenu Element
- Changes the way Progress Circle calculates its size and SVG position for it to blend in seamlessly with Material Icons.


src/components/IconButton/index.js
```diff
@@ -0,0 +1,123 @@
...
+  hover: css({
+    '@media(hover)': {
+      ':hover > *': {
+        opacity: 0.6
+      }
+    }
+  }),
+  label: css({
+    fontFamily: fontFamilies.sansSerifMedium,
+    fontSize: 12,
```
tpreusse, line 105:
```suggestion
    fontSize: 15,
```
Im Zweifelsfall für grössere Schriften. Will keine dickere Brille ;-)

src/components/Discussion/Internal/Comment/Header.js
```diff
@@ -268,7 +269,11 @@ export const Header = ({ t, comment, menu, isExpanded, onToggle }) => {
           {isExpanded ? <IcCollapse /> : <IcExpand />}
         </button>
       )}
-      {menu && <CalloutMenu align='right'>{menu}</CalloutMenu>}
+      {menu && (
+        <CalloutMenu Element={props => <MoreIcon {...props} />} align='right'>
```
tpreusse, line 273:
```suggestion
        <CalloutMenu Element={MoreIcon} align='right'>
```
This pattern should be avoided. And seems unnecessary here.

`props => <MoreIcon {...props} />` creates a new function on every render. React uses function identity internally and if it is constantly changing it will cause unnecessary complex DOM re-building.

If it were necessary the component should be defined outside of the render method:
```
const MoreIconWithProp = props => <MoreIcon size={24} {...props} />
const MyComponent = () => {
  return <CalloutMenu Element={MoreIconWithProp} />
}
const MyVeryDynamicComponent = ({size}) => {
  const MoreIconWithDynamicProp = useMemo(() => props => <MoreIcon size={size} {...props} />, [size])
  return <CalloutMenu Element={MoreIconWithDynamicProp} />
}
```

src/components/TeaserFeed/Container.js
```diff
@@ -47,7 +48,12 @@ const Teaser = ({
             float: 'right'
           }}
         >
-          <CalloutMenu align='right'>{menu}</CalloutMenu>
+          <CalloutMenu
+            Element={props => <MoreIcon width='calc(1em + 7px)' {...props} />}
```
tpreusse, line 52: see other comment about not defining react components on render

tpreusse (changes requested):
Looking good. I've noticed an issue with dynamic React components and the `font-size` is still a bit of an issue. Let's post it in #design to resolve the 2nd one and let the channel do it's magic and decide.

src/components/IconButton/index.js
ovbm: Wieso 15? 14 würde doch auch reichen...

tpreusse: 14 ist für mich auch ok

src/components/IconButton/index.js
```diff
@@ -0,0 +1,123 @@
...
+  ) => {
+    const Element = href ? 'a' : 'button'
+    const customStyles = style || null
+    const [colorScheme] = useColorContext()
+
+    return (
+      <Element
+        {...styles.button}
+        {...((onClick || href) && styles.hover)}
+        style={{ cursor: !onClick ? 'auto' : 'pointer', ...customStyles }}
```
tpreusse, line 34:
```suggestion
        style={{ cursor: href || onClick ? 'pointer' : 'auto', ...customStyles }}
```
Currently external links without an `onClick` e.g. on profile, do not get pointers.

src/components/IconButton/index.js
```diff
@@ -0,0 +1,123 @@
...
+  }),
+  hover: css({
+    '@media(hover)': {
+      ':hover > *': {
+        opacity: 0.6
+      }
+    }
+  }),
+  label: css({
+    fontFamily: fontFamilies.sansSerifMedium,
```
tpreusse, line 104: Oh I've overlooked a problem:
> The old fontFamilies export is deprecated and should no longer be used.
> [Typography](https://styleguide.republik.ch/typography)

Use `fontStyles` instead:

```suggestion
    ...fontStyles.sansSerifMedium,
```

tpreusse (approved):
I previously overlooked an issue re font styles (causing medium to not always being applied, e.g. in styleguide docs)

Otherwise approval from my side!

src/components/IconButton/index.js
```diff
@@ -101,7 +104,7 @@ const styles = {
     }
   }),
   label: css({
-    fontFamily: fontFamilies.sansSerifMedium,
+    fontFamily: fontStyles.sansSerifMedium,
```
tpreusse, line 107: you need to spread it:
```suggestion
    ...fontStyles.sansSerifMedium,
```
(contains family and weight)

PR #317, opened 2020-09-03
Increase margin on IconButton

ovbm:
Right before the release of the action bar, I changed the IconButton default margin from 20px to 18px. I think it's a bit too narrow and on small mobile phones users could benefit from a bit more space, so I reverted it back to 20px. See narrowest scenario below iPhone 5S (it's subtle): 

after:
Screenshot 2020-09-03 at 08 16 11

before
Screenshot 2020-09-03 at 08 16 29


tpreusse (approved):
FAT FINGERS FTW

PR #322, opened 2020-09-23
Dark Mode Images

annatraussnig:
Adjusted the `Article` & `Newletter` schema to parse a supplementary image. We need to discuss which schemas make sense to support

### TODOS

- [ ] Support more schemas if necessary
- [ ] `<Figure />` UI
- [x] Match DM PR logic to conditionally show/hide images


tpreusse (approved):
Looks good. I added a `SwitchImage` logic, `dark` `srcSet` support and some docs.

src/components/Figure/Image.js
```diff
@@ -46,78 +46,88 @@ const GalleryButton = ({ gallerySize, onClick }) => {
...
-      : () => {}
+  const onClick =
+    enableGallery && context.toggleGallery
+      ? () => context.toggleGallery(src)
+      : undefined
 
-    const size = sizeProp || (sizeProp === undefined && imageSizeInfo(src))
-    const aspectRatio = size ? size.width / size.height : undefined
+  const size = sizeProp || (sizeProp === undefined && imageSizeInfo(src))
+  const aspectRatio = size ? size.width / size.height : undefined
```
tpreusse, line 69: Because it already happened during testing, let's ensure we always pick the smaller and therefore higher `aspectRatio`:
```suggestion
  let aspectRatio = size ? size.width / size.height : undefined
  if (dark?.size) {
    aspectRatio = Math.min(aspectRatio, dark.size.width / dark.size.height)
  }
```

(we still don't support it and should leave the note in place but prevent overlap at the same time)

PR #330, opened 2020-10-23
Dynamic component identifiers

annatraussnig:
- Add an `identifier` prop to `DynamicComponent` as an alternative to `src`
- Add a `dynamicComponentIdentifiers` dictionary to `createArticleSchema` used to map identifiers and components in the frontend
- Document the new properties


src/templates/Article/dynamicComponent.js
```diff
@@ -39,10 +40,17 @@ const createDynamicComponent = ({
...
       html: html && html.value,
       props: node.data.props,
       loader: node.data.loader,
-      require: dynamicComponentRequire
+      require: dynamicComponentRequire,
+      identifiers:
+        dynamicComponentIdentifiers ||
+        (node.data.identifier && {
+          [node.data.identifier]: () => null
+        }) ||
```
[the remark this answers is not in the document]
tpreusse: Ah was it in the Publikator preview? I just thought about that case some more and would recommend using a getter in Publikator instead. I will comment in the related PR.

tpreusse (commented):
Generally looks good but the default null fallback should be removed.

src/components/DynamicComponent/index.js
```diff
@@ -36,31 +36,38 @@ class DynamicComponent extends Component {
...
-      })
+        .catch(error => {
+          this.setState({ error })
+        })
+    }
   }
   render() {
     const { LoadedComponent } = this.state
-
+    const IdentifierComponent = this.props.identifiers[this.props.identifier]
```
tpreusse, line 56: One more thing, let's ensure they are mutually exclusive:
```suggestion
    const IdentifierComponent = !this.props.src && this.props.identifiers[this.props.identifier]
```

src/components/DynamicComponent/index.js
```diff
@@ -36,31 +36,38 @@ class DynamicComponent extends Component {
...
       throw error
     }
 
-    if (LoadedComponent) {
+    const DisplayComponent = LoadedComponent || IdentifierComponent
+
+    if (DisplayComponent) {
       return (
-        <LoadedComponent require={this.props.require} {...this.props.props} />
+        <DisplayComponent require={this.props.require} {...this.props.props} />
```
tpreusse, line 70: To make out life easier for placeholders and such: let's always add the `identifier` as prop:
```suggestion
        <DisplayComponent require={this.props.require} identifier={this.props.identifier} {...this.props.props} />
```

tpreusse (approved):

PR #333, opened 2020-10-28
Darkmode Fixes

ovbm:
Small fixes for darkmode:
- Email Editorial links inherit color
- Interaction.A as alternative for linkRule // UNNECESSARY, A is already exported from Typography. needs to be removed.
- Scribble titles now darkmode ready


annatraussnig (approved):
🌕 🌔 🌑

src/templates/EditorialNewsletter/web/index.js
```diff
@@ -34,6 +34,7 @@ const createSchema = ({ ...args } = {}) => {
     List,
     ListItem,
     ListP,
+    Link: A,
```
ovbm, line 37: Using this causes the app to crash: `Error: React.Children.only expected to receive a single React element child.`
Any Idea as to why? @tpreusse

src/components/Typography/Editorial.js
```diff
@@ -265,16 +265,26 @@ export const link = css({
...
   const [colorScheme] = useColorContext()
-  const colors = css({
-    color: colorScheme.text,
-    '@media (hover)': {
-      ':hover': {
-        color: colorScheme.lightText
-      }
-    }
-  })
+  const AHoverRule = useMemo(
```
tpreusse, line 268:
```suggestion
  const hoverRule = useMemo(
```

Variables starting with capital letters always smell of React components ;-)

tpreusse (approved):

PR #345, opened 2020-11-19
Dark Mode Round 99

tpreusse:
- feat(colors): ColorContextLocalExtension
- feat(Charts): custom dark mode colors via colorDarkMapping
- fix(Charts): add missing color mappings for dark mode
- refactor(colors): clean naming, always use `set` or `getCSSColor`

`ColorContextLocalExtension` allows to define new variables for a specific local react tree.

Based on that support for `colorDarkMapping` in chart configs was added. Allowing to re-map certain bright colors to dark colors when in dark mode. Not sure about naming yet. 

Screenshot 2020-11-19 at 23 53 23

Also `colorScheme.text` et al are no longer available.


tpreusse:
Merging because I want to release tomorrow morning 🙈

PR #348, opened 2020-11-23
fix(IconButton): display labels in IE11

tpreusse:
see https://caniuse.com/css-initial-value

I don't see any benefit of using initial here but maybe I'm overseeing something.

Screenshot 2020-11-23 at 15 45 28
Screenshot 2020-11-23 at 15 48 07

### With explicit inline

Screenshot 2020-11-23 at 15 45 38

Screenshot 2020-11-23 at 15 48 50

### DIFF
--- a/src/components/IconButton/index.js
+++ b/src/components/IconButton/index.js
@@ -115,11 +115,11 @@ const styles = {
   long: css({
     display: 'none',
     [mUp]: {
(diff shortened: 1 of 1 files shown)


ovbm:
don't know why it was 'initial' in the first place...

ovbm (approved):

PR #350, opened 2020-11-30
Comment with preview

annatraussnig:
Screenshot 2020-11-30 at 17 46 52

### Also in this PR

- dark mode for active debates teaser

TODO: 

- [x] fix broken links
- [x] tweak workflow to feature comments in FE/BE

### DIFF
--- a/src/components/CommentTeaser/CommentTeaser.js
+++ b/src/components/CommentTeaser/CommentTeaser.js
@@ -100,7 +100,6 @@ export const CommentTeaser = ({
   const isDesktop = useMediaQuery(mUp)
   const [colorScheme] = useColorContext()
   const highlight = get(highlights, '[0].fragments[0]', '').trim()
-  const documentPreview = discussion?.document?.meta?.twitterImage
   const commentCount = discussion?.comments?.totalCount
 
   const endsWithPunctuation =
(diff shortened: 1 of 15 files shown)


src/components/CommentTeaser/CommentTeaser.js
```diff
@@ -182,16 +195,49 @@ export const CommentTeaser = ({
             </Link>
           </CommentBodyParagraph>
         </div>
+
+        {documentPreview && (
```
ovbm, line 199: Question: This changes all comment teasers to add an image if they have a twitter image. But in an element like the discussion section on the main feed, we wouldn't want images there. Should there be a prop "withPreview"? Or does the query there not ask for "document meta twitterImage" and thus it's ok to not use a prop?
Screenshot 2020-12-04 at 11 08 14

src/components/Discussion/Internal/Comment/IconLink.js
```diff
@@ -62,18 +56,34 @@ export const IconLink = ({
...
-    <a href={href} onClick={onClick} {...styles.link} style={patchedStyle}>
+    <a
+      href={href}
+      onClick={onClick}
+      {...styles.link}
+      {...linkStyleRule}
+      style={patchedStyle}
+    >
       <span {...styles.icon}>
-        <Icon size={dimension} {...colorScheme.set('fill', 'primary')} />
```
ovbm, line 69: why remove fill here?

annatraussnig: The color is now set through the `linkStyleRule`. It includes a hover state and fixes the hover color, which was different from the other links.

src/components/CommentTeaser/CommentTeaser.js
annatraussnig: You make a good point. Using the twitter image implicitly is weird.

src/components/Discussion/Internal/Comment/IconLink.js
ovbm: 👍🏼

ovbm (approved):

PR #352, opened 2020-12-04
fix(inQuotes): fix replaceAll not supported in FF

annatraussnig:
my bad…

### DIFF
--- a/src/lib/inQuotes.js
+++ b/src/lib/inQuotes.js
@@ -14,7 +14,7 @@ export const inQuotes = (str, marks = defaultMarks) => {
     return quotedStr
   }
   quotedStr = quotedStr
-    .replaceAll(marks.outerOpening, marks.innerOpening)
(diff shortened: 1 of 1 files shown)


ovbm:
I didn't know that XD

ovbm (approved):

annatraussnig:
No wait, it is https://caniuse.com/?search=replaceAll… it was working in the SG docs, but not in server side rendering in the frontend (got `replaceAll is not a function`). Maybe @tpreusse knows something I don't.

tpreusse:
@annatraussnig not supported in (our) node js version yet:
```
$ node
Welcome to Node.js v14.4.0.
Type ".help" for more information.
> ''.replace
[Function: replace]
> ''.replaceAll
undefined
> 
```
But would also be a problem for older browsers, even older Chromes and Mobile Safaris.

PR #355, opened 2020-12-15
Discussion: Hide unpublished nodes

patrickvenetz:
Comments can be unpublished. Header will replace author with a label.

Bildschirmfoto 2020-12-16 um 12 52 47

If comment was unpublished by admin (`adminUnpublished`), label will indicate that.

Bildschirmfoto 2020-12-16 um 12 52 53

### DIFF
--- a/src/components/Discussion/Internal/Comment/Header.js
+++ b/src/components/Discussion/Internal/Comment/Header.js
@@ -163,8 +163,8 @@ export const Header = ({ t, comment, menu, isExpanded, onToggle }) => {
     displayAuthor,
     updatedAt,
     createdAt,
-    published,
-    adminUnpublished,
+    published = true,
+    adminUnpublished = false,
     comments,
     parentIds = []
   } = comment
--- a/src/components/Discussion/Internal/Comment/Header.js
+++ b/src/components/Discussion/Internal/Comment/Header.js
@@ -163,6 +163,8 @@ export const Header = ({ t, comment, menu, isExpanded, onToggle }) => {
     displayAuthor,
     updatedAt,
     createdAt,
+    published,
+    adminUnpublished,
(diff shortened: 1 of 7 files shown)


src/components/Discussion/Internal/Comment/Header.js
```diff
@@ -214,43 +216,54 @@ export const Header = ({ t, comment, menu, isExpanded, onToggle }) => {
         }
       })()}
       <div {...styles.center}>
-        <Link displayAuthor={displayAuthor} passHref>
-          <a {...styles.name} {...colorScheme.set('color', 'text')}>
-            {name}
-          </a>
-        </Link>
+        {!published && (
```
ovbm, line 219: detail, why not:
`{ published ? (<PublishedStuff/>) : (<UnpublishedStuff/>)}`

ovbm (approved):

src/components/Discussion/Internal/Comment/Header.js
patrickvenetz: Thought so, too.

Yet, pattern spread all over these files is usually `foobar && <Something />`, `!foobar && <SomethingElse />` and was inclined to stick with it.

I assume it has some advantages, when …

… condition for truthy changes but not for falsy wouldn't e.g.
```
foobar && buzzfizz && <Something />
!foobar && <SomethingElse />
```

… nested conditions e.g.

```
foobar && !fizzbuzz && <Something />
foobar && fizzbuzz && <SomethingFizzy />
!foobar && <SomethingElse />
```

(instead of)

```
foobar ? (
  fizzbuzz ? (
    <Something />
  ) : (
    <SomethingFizzy />
  )
) : (
  <SomethingElse />
)
```

PR #356, opened 2020-12-15
Enables setting mobile padding of callout contents as props

ovbm:
This huge whitespace is only necessary in iOS non-native safari. It drives me nuts (also, i reduced it to 50px from 80px).
IMG_2159

### DIFF
--- a/src/components/Form/Field.js
+++ b/src/components/Form/Field.js
@@ -41,17 +41,7 @@ const styles = {
     borderRadius: 0,
     backgroundColor: 'transparent'
   }),
-  codeField: css({
-    fontFamily: 'Courier, Courier New, monospace',
-    fontSize: 14
-  }),
   labelText: css({
-    position: 'absolute',
-    left: X_PADDING,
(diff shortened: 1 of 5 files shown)


src/components/Form/Field.js
```diff
@@ -41,12 +41,22 @@ const styles = {
...
     left: X_PADDING,
     top: LINE_HEIGHT + Y_PADDING,
     transition: 'top 200ms, font-size 200ms'
   }),
+  labelTextCode: css({
+    position: 'absolute',
+    left: X_PADDING,
+    top: LINE_HEIGHT + Y_PADDING,
+    transition: 'top 200ms, font-size 200ms'
+  }),
```
annatraussnig, line 59: isn't this already deployed?

annatraussnig (approved):

src/components/Form/Field.js
ovbm: good catch, i should delete it. I had to overwrite it in publikator itself.

PR #361, opened 2021-01-27
Darkmode 1.2

annatraussnig:


src/components/Callout/index.js
```diff
@@ -90,7 +90,7 @@ const Callout = ({
       css({
         backgroundColor: colorScheme.getCSSColor('overlay'),
         [mUp]: {
-          boxShadow: colorScheme.getCSSColor('overlayShadow')
+          boxShadow: `inset 0px -50px 50px -30px colorScheme.getCSSColor('overlayShadow')`
```
ovbm, line 93: Alignment is off:
Screenshot 2021-01-27 at 14 44 40

src/components/Logo/BrandMark.js
```diff
@@ -40,16 +41,19 @@ const styles = {
...
-R.defaultProps = {
-  fill: '#000'
+const R = ({ fill }) => {
+  const [colorScheme] = useColorContext()
+  return (
+    <div {...styles.container}>
+      <svg
+        {...styles.svg}
+        viewBox={VIEWBOX}
+        {...colorScheme.set('fill', 'logo')}
```
ovbm, line 51: why both a svg fill and a path fill? And why is only the svg from props?

tpreusse: the path fill allow overwrite to get a specific color, IMO can be kept. Setting the fill attr on the svg would not work because the css is more specific if on the same element

tpreusse: one alternative would be:
```suggestion
        {...colorScheme.set('fill', fill || 'logo')}
```
an then remove the fill on the path.

src/components/CommentTeaser/CommentTeaser.js
```diff
@@ -71,7 +72,10 @@ const styles = {
   image: css({
     borderBottomWidth: 1,
     borderBottomStyle: 'solid',
-    width: '100%'
+    width: '100%',
+    [mUp]: {
+      marginTop: 20
```
tpreusse, line 77: look weird / leads to double margin because `imageContainer` already has 20px for mUp
Screenshot 2021-02-03 at 00 13 08

src/components/Form/Checkbox.js
```diff
@@ -25,15 +25,15 @@ const Checkbox = ({
     : disabled
     ? colorScheme.set('borderColor', 'disabled')
     : black
-    ? colorScheme.set('borderColor', '#000')
+    ? colorScheme.set('borderColor', 'logo')
```
tpreusse, line 28: just out of curiosity: where is Checkbox still used with `black` that lead to this change? We started to get ride of that in `Field`. And should probably continue soon with checkbox (not as part of this PR).

src/components/TeaserFront/Tile.js
```diff
@@ -106,58 +108,60 @@ const Tile = ({
...
-          <LazyLoad
-            visible={aboveTheFold}
-            consistentPlaceholder
-            type='span'
-            style={{
-              position: 'relative',
-              fontSize: 0,
-              display: 'inline-block'
-            }}
+    <ColorContextProvider colorSchemeKey={colorSchemeKey}>
```
tpreusse, line 111: this is a bit suboptimal and may not lead to the desired effects: concretely the `colorScheme.set('backgroundColor', 'default')` just underneath will not actually be based on this context but the one above since `useColorContext` is called beforehand.

Could we move the `ColorContextProvider` to `src/templates/Front/index.js` with a hardcoded `light`? A additional benefit of that approach would also be that it will not have a negative performance effect of repeating a auto context when not necessary (repeating / nested context with a `auto` value may not work as expected in many situations).

src/components/Form/Checkbox.js
annatraussnig: ah that was in a legacy page `/umfrage/1-minute` – I thought it might come up again in the future so I decided to fix it. But if you are telling me this is soon to be deprecated, I can revert that change. Should I?

tpreusse: thx. leave it as is, just wanted to know the culprit page

tpreusse (approved):

PR #365, opened 2021-02-10
feat(TileRow): adds noPadding prop for marketing front

ovbm:
Required for even spacing of elements on Marketing-Front

before
Screenshot 2021-02-10 at 16 05 00

after
Screenshot 2021-02-10 at 16 03 12

### DIFF
--- a/src/components/TeaserFront/TileRow.js
+++ b/src/components/TeaserFront/TileRow.js
@@ -133,6 +133,14 @@ const styles = {
         ...sizeSmall
       }
     }
+  }),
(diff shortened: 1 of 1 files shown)


tpreusse:
I don't see the point of using `TeaserFrontTileRow` and `TeaserFrontTile` around `CommentTeaser`. `TeaserFrontTileRow` serve a specific role and are not just generic grid components. Let's discuss tomorrow.

ovbm:
yup, makes sense

PR #367, opened 2021-02-12
fix(BarChart): keep dark mode inline labels same as light mode

annatraussnig:


src/components/Chart/utils.js
```diff
@@ -252,5 +253,5 @@ export const unsafeDatumFn = code => new Function('datum', `return ${code}`)
 export const getTextColor = bgColor => {
   const color = rgb(bgColor)
   const yiq = (color.r * 299 + color.g * 587 + color.b * 114) / 1000
-  return yiq >= 128 ? 'text' : 'textInverted'
+  return yiq >= 128 ? colors.light.text : colors.light.textInverted
```
tpreusse, line 256: Let's go back to the original version with true black and white for maximum contrast:
```suggestion
  return yiq >= 128 ? 'black' : 'white'
```
Source: https://github.com/orbiting/styleguide/commit/e5d85f81b0d9917f645cfecdea73f3a5577a392d#diff-e502f82cb111bef7174ae02cba44116da3eff0ef79eed8c8c9442be5c089ad9fR238-R242

tpreusse (approved):

PR #369, opened 2021-02-22
Share Tafel Generator (frontend)

ovbm:
- [x] move Logo in Brand


ovbm (commented):
Looks good IMO (can't approve because I started pr)

src/components/ShareImage/ShareImagePreview.js
```diff
@@ -0,0 +1,138 @@
...
+          (shareImage && shareImageJustify[textPosition]) || 'center',
+        borderWidth: embedPreview ? 2 : 0
+      }}
+    >
+      {format?.image && <img {...styles.formatImage} src={format?.image} />}
+      {format?.title && (
+        <div
+          {...styles.formatTitle}
+          style={{
+            color: coloredBackground ? '#FFF' : format?.color,
```
annatraussnig, line 116: Color can also come from format kind -> needs to be added
`const formatColor = colorMeta && (colorMeta.color || colors[colorMeta.kind])`

tpreusse (commented):
I really like the editing experience in Publikator. Here a few code comments. I think we should also do a naming iteration: see [backend review](https://github.com/orbiting/backends/pull/577#pullrequestreview-634020586).

src/components/ShareImage/docs.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+# Share Image Generator
+
+A `<ShareImageGenerator />` takes a format type and returns a form to create a share image for that specific format type style.
+
+Supported props:
+
+- `socialKey`: required. Either `twitter` or `facebook`
+- `data` and `onInputChange`: an immutable `Map` and a setter. Both based on meta data model from Publikator
```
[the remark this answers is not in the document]
annatraussnig: I see. Pfffff... Time to move some code around then. 😄

src/components/ShareImage/docs.md
```diff
@@ -0,0 +1,236 @@
...
+        title: 'Briefings' 
+      }
+    },
+    color: '#00809C'
+  }}
+/>
+```
+
+## Dialog
+Setting type to 'Dialog' enables to choose between Sans-Serif and Serif fonts.
```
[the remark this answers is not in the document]
annatraussnig: I leave that question to @ovbm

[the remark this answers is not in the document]
ovbm: There seems to be a difference between the sketch file and the actual share images they are using... In the sketch file they use either sans-serif or serif. If in practice they only ever use meta, we can remove this.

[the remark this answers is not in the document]
tpreusse: Let's remove it for now and re-evaluate if it comes up in practice. Same for meta, where we already removed it but the docs need an update:
Screenshot 2021-04-14 at 10 44 02
(Text talks about font switch but it's already removed)

src/components/ShareImage/docs.md
```diff
@@ -0,0 +1,394 @@
...
+  inverted: false,
+  text: '',
+  format: { 
+    title: 'Aus der Arena',
+    section: {
+      meta: {
+        title: 'Format' 
+      }
+    },
+    kind: 'scribble'
```
ovbm, line 201:
```suggestion
    kind: 'scribble',
    color: '#D74132'
```

ovbm: needs a color so that background is correct afaik

annatraussnig: does it? we use 
`const formatColor = format?.color || colors[format?.kind]`
to determine the color.

ovbm: ok, so that doesn't work in this case. It's a black background without a specific color

annatraussnig: aaaah for the background color. Got it

ovbm: what if we do 
`  const formatBackgroundColor = format?.color || colors[format?.kind]
`
and
`        backgroundColor: inverted ? formatBackgroundColor : '#FFF',
`

ovbm: ah forget that

ovbm: formatColor is the way to go, thx!

annatraussnig: pushed a fix

ovbm:
LGTM

Issue #312, opened 2020-07-13
Rewind & Skip Buttons for Audio-Player


tpreusse:
Allow to rewind 15 seconds and skip 30 seconds like Apple Podcasts and others.

- [Source Input](https://www.republik.ch/dialog?t=general&focus=f4c969d8-96a5-4449-82c7-715fd4248d44)

PR #372, opened 2021-03-14
add rewind & skip buttons

panmau:
This pull request is a first shot at addressing issue #312.

I added forward and backward buttons to the player that skip forward/backward 30 seconds in the audio file when clicked. 
10 or 5 seconds would be the alternative to skipping 30 seconds as these icons are also available in the Material Iconset.

This is how the new buttons look (screenshot taken from the styleguide):

Some remaining questions from my side:
Should we make the button display optional? (as with the download button)
And the mobile view seems a bit crowded with buttons now: should we display them on mobile?

Would love to hear your thoughts about my first PR in one of the orbiting Repos.

### DIFF
--- a/src/components/AudioPlayer/Player.js
+++ b/src/components/AudioPlayer/Player.js
@@ -6,6 +6,8 @@ import Pause from 'react-icons/lib/md/pause'
 import Rewind from 'react-icons/lib/md/skip-previous'
 import Close from 'react-icons/lib/md/close'
 import Download from 'react-icons/lib/md/file-download'
+import Forward from 'react-icons/lib/md/forward-30'
+import Replay from 'react-icons/lib/md/replay-30'
 
 import { ellipsize } from '../../lib/styleMixins'
 import { timeFormat } from '../../lib/timeFormat'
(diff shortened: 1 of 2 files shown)


tpreusse:
Thank you Maurice.

Awesome stuff.

Already looks pretty good, I'd also say that it needs some tweaks for mobile. The native audio player of the app has the following layout:
IMG_4191

I'd suggest to go that route to and would get ride of the rewind—`setTime(0)`—button, I don't think that it is widely used and can be accomplished via dragging the nob too. I'll check with a colleague tomorrow and we'll get back to you.

maracuja-juice:
Thank you Thomas for the first feedback.

I suggest that we change the button layout on all screen sizes to the layout in the app. Also I agree that the rewind (`setTime(0)`) button can be removed as it isn't that useful.
Additionally I will replace the 30 seconds backward button with a 10 second backward button. I wasn't aware that this is already implemented in the app, as I haven't used it yet.

I think that with these changes the usability will be improved on all screen sizes.

maracuja-juice:
I updated the PR so that the button layout matches the app layout and the replay button only moves 10 seconds backwards instead of 30. 

This is how it looks now:

ovbm (approved):
LGTM

Issue #362, opened 2021-02-03
Discussion Composer draft prefill


patrickvenetz:
Discussion composer saves drafted, unsubmitted answers and restores them when you return to page. Composer seems not to keep track to which comment a drafted answer belongs to.

While restoring a draft is handy indeed, it's current implementation lead to [confusion](https://www.republik.ch/dialog?t=article&id=0a844046-6b50-4124-9646-dac63879d4ff&focus=f2ba0cf6-be39-4a89-aba9-8ce6083289f3) [multiple](https://www.republik.ch/dialog?t=article&id=0ceb3de2-b796-4061-9650-a65f399ab4a4&focus=4a1bfd05-6149-4876-8ca5-7e74decbc065) [times](https://www.republik.ch/2021/02/01/ein-jahr-corona-in-der-republik-waren-wir-brauchbar-fuer-sie/diskussion?focus=717e3ce9-9264-46df-85b9-118981991203) on republik.ch (linked are some examples). Predominant mishap seems posting a drafted answer as a new root comment.

https://user-images.githubusercontent.com/331800/106734587-7a779080-6613-11eb-89ef-914dc81d8b43.mov

**Steps to reproduce in republik.ch**
1. Start typing answer to a thread but do not submit
2. Navigate to another page
3. Return to discussion and find drafted answer in all composer fields

patrickvenetz:
![](https://media0.giphy.com/media/ar4x1w44umngk/giphy.gif?cid=ecf05e47znuiegn6xxw0rkoo9ibfsmngozv4529x52spad3t&rid=giphy.gif&ct=g)

PR #375, opened 2021-03-21
improve Discussion Composer prefill experience

panmau:
This is my proposal on how to improve the prefill experience in the Discussion Composer. (#362)
Instead of only saving the discussionId, the id of the comment the user wants to reply to is saved aswell. The draft is only restored when both ids match.
For root comments, the behaviour is not changed.

I tested if this new behaviour works well within the republik frontend by locally building the library as a package and passing  the parentId prop to the key function.

This is how it looks in action:

https://user-images.githubusercontent.com/16801528/111913132-0aab5100-8a6d-11eb-8f87-f04cc5079ca1.mp4

This change will probably cause confusion in the first few days/weeks after the change is introduced. Because the drafts that are currently saved as replies to comments, won't be restored. But they won't be lost, as they can still be restored in the root comment field. 
I don't see something we can do to prevent this though, as there's no way we can migrate the data. 
But if it is well communicated in a short Update article, I think this confusion can be avoided almost completely. 
Additionally in the long term it will cause less confusion which in my opinion totally outweighs the possibly caused confusion in the short term.

### DIFF
--- a/src/components/Discussion/Composer/CommentComposer.js
+++ b/src/components/Discussion/Composer/CommentComposer.js
@@ -55,8 +55,10 @@ const styles = {
  * particular, this allows the frontend to directly display the CommentComposer
  * if there is text stored in localStorage.
  */
-export const commentComposerStorageKey = discussionId =>
-  `commentComposerText:${discussionId}`
+export const commentComposerStorageKey = (discussionId, parentId) => {
+  const key = `commentComposerText:${discussionId}`
+  return parentId ? key + `:${parentId}` : key
+}
 
 export const CommentComposer = props => {
   const {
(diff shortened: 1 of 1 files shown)


tpreusse:
Hello Maurice

First of all thank you for the contribution. Second, sorry for delay in reviewing.

I worry a bit that this approach creates many abandoned local storage keys and users having a hard time finding their draft texts.

Currently the root composer is automatically set to active in [the frontend](https://github.com/orbiting/republik-frontend/blob/deb6befe253caffd25290ae29e65537992838824/components/Discussion/DiscussionCommentComposer.js#L64-L66) when there is something in the local storage.

I'd prefer an approach where we'd save all drafts for an discussion in one local storage key as a json object, an example structure:
```json
{
  "root": "Love that text.",
  "answers": [
    {"parentId": "327b9b6b-3fb7-44b1-a4fe-ffb42c5ec93a", "text": "You are right."},
    {"parentId": "6ef1ef25-91b8-44eb-815d-169d2d737a92", "text": "No, no, no!"}
  ]
}
```
(Would require a small utility to modify this object in the local storage without spreading complexity, probably on a new key space and with automatic migration of the old key space for the first month or so.)

With that object we could print a list of drafts underneath the root composer with links that jump to the parent id.

Additionally we should also add an automatic reply composer opening logic when a comment is rendered that has an answer draft.

Does that make sense? Makes the task quite a bit more complex but I believe it will lead to a truly better user experience.

PR #376, opened 2021-03-26
fix(BarChart): fix text color when bar is smaller than label

ovbm:
before
Screenshot 2021-03-26 at 14 42 50

after
Screenshot 2021-03-26 at 14 40 48

### DIFF
--- a/src/components/Chart/Bars.js
+++ b/src/components/Chart/Bars.js
@@ -10,6 +10,7 @@ import { fontFamilies } from '../../theme/fonts'
 import { underline } from '../../lib/styleMixins'
 import { useColorContext } from '../Colors/useColorContext'
 
+import { sansSerifRegular12 as LABEL_FONT } from '../Typography/styles'
 import {
   calculateAxis,
   groupBy,
@@ -21,6 +22,7 @@ import {
   getTextColor
 } from './utils'
 import ColorLegend from './ColorLegend'
+import { createTextGauger } from '../../lib/textGauger'
 
 import { getColorMapper } from './colorMaps'
 
@@ -123,6 +125,11 @@ const styles = {
     borderRadius: '4px'
   })
 }
+
+const labelGauger = createTextGauger(LABEL_FONT, {
+  dimension: 'width',
+  html: true
+})
 
 const BarChart = props => {
   const {
@@ -442,14 +449,28 @@ const BarChart = props => {
                           ? 'left'
                           : 'right')
                       let iTextAnchor = 'middle'
+
+                      const inlineLabelText = [
+                        inlineValue && xAxis.format(segment.value),
+                        inlineValueUnit && inlineValueUnit,
+                        inlineLabel && segment.datum[inlineLabel]
+                      ].join(' ')
+                      const inlineLabelTextWidth = labelGauger(inlineLabelText)
+
                       let iXOffset = segment.width / 2
                       if (inlinePos === 'right') {
                         iTextAnchor = 'end'
                         iXOffset = segment.width - 5
+                        if (segment.width <= inlineLabelTextWidth) {
+                          iXOffset = -segment.width
+                        }
                       }
                       if (inlinePos === 'left') {
                         iTextAnchor = 'start'
                         iXOffset = 5
+                        if (segment.width <= inlineLabelTextWidth) {
+                          iXOffset = 5 + segment.width
+                        }
                       }
 
                       return (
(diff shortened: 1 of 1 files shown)


annatraussnig (approved):

annatraussnig:
Looks good. Quick question: why so much left padding here? `-200 c`
Screenshot 2021-03-26 at 15 10 49

annatraussnig:
It's very much a detail though – don't let it hold you back from merging. 🎉

ovbm:
good question. I also don't really understand why some of the inlinelabels in this graph are white and others black...

annatraussnig (approved):
LGTM!

ovbm:
@tpreusse tested it locally - should be fine with squash and merge.

src/components/Chart/Bars.js
```diff
@@ -329,12 +336,28 @@ const BarChart = props => {
...
             : 'right')
+
+        d.inlineLabel = [
+          inlineValue && xAxis.format(d.value),
+          inlineValueUnit && inlineValueUnit,
+          inlineLabel && d.datum[inlineLabel]
+        ].join(' ')
+        d.inlineLabelTextWidth = labelGauger(d.inlineLabelText)
+
+        const shiftInlineText = isLast && d.width <= d.inlineLabelTextWidth
```
tpreusse, line 347:
```suggestion
        d.shiftInlineText = bar.segments.length === 1 && d.width <= d.inlineLabelTextWidth
```
safe on segment for later use in rendering (ensuring the different color rule is only applied when shifted)

And restrict to single segment. Multi segment has bad edge cases even when last:
Screenshot 2021-04-09 at 00 16 47

src/components/Chart/Bars.js
```diff
@@ -329,12 +336,28 @@ const BarChart = props => {
...
+
         if (d.inlinePos === 'right') {
           d.iTextAnchor = 'end'
           d.iXOffset = d.width - 5
+          if (shiftInlineText) {
```
tpreusse, line 352:
```suggestion
          if (d.shiftInlineText) {
```

src/components/Chart/Bars.js
```diff
@@ -329,12 +336,28 @@ const BarChart = props => {
...
+            d.iXOffset = -5
+          }
         } else if (d.inlinePos === 'left') {
           d.iTextAnchor = 'start'
           d.iXOffset = 5
+          if (shiftInlineText) {
```
tpreusse, line 358:
```suggestion
          if (d.shiftInlineText) {
```

src/components/Chart/Bars.js
```diff
@@ -464,16 +487,15 @@ const BarChart = props => {
                             y={bar.style.inlineTop}
                             dy='1em'
                             fontSize={bar.style.fontSize}
-                            fill={getTextColor(segment.color)}
+                            fill={
+                              segment.width >= segment.inlineLabelTextWidth &&
+                              getTextColor(segment.color)
+                            }
+                            {...(segment.width < segment.inlineLabelTextWidth &&
+                              colorScheme.set('fill', 'text'))}
```
tpreusse, line 495:
```suggestion
                            {...colorScheme.set('fill', segment.shiftInlineText ? 'text' : getTextColor(segment.color))}
```

src/components/Chart/Bars.js
```diff
@@ -482,7 +504,13 @@ const BarChart = props => {
...
                               dy='1em'
                               fontSize={bar.style.secondaryFontSize}
-                              fill={getTextColor(segment.color)}
+                              fill={
+                                segment.width >= segment.inlineLabelTextWidth &&
+                                getTextColor(segment.color)
+                              }
+                              {...(segment.width <=
+                                segment.inlineLabelTextWidth &&
+                                colorScheme.set('fill', 'text'))}
```
tpreusse, line 513:
```suggestion
                              {...colorScheme.set('fill', segment.shiftInlineText ? 'text' : getTextColor(segment.color))}
```

tpreusse (changes requested):

src/components/Chart/Bars.js
ovbm: This would negate the fix in some cases see:
Screenshot 2021-04-09 at 08 50 59
I think in this case it's actually a configuration issue. The bars can be configured to look right with (isLast)
Screenshot 2021-04-09 at 08 59 23

tpreusse: @ovbm found a solution which automatically does it:
https://github.com/orbiting/styleguide/pull/376/commits/ae32178ff3afb84753b876ee2457082be17c565e

Hopefully it does not have any regression.

tpreusse (approved):
Good to ship from my side.

ovbm:
nice thx!

PR #382, opened 2021-04-09
fix(Barchart): pass empty string to lablegauger if no string

ovbm:
prevents error in frontend

### DIFF
--- a/src/components/Chart/Bars.js
+++ b/src/components/Chart/Bars.js
@@ -333,7 +333,7 @@ const BarChart = props => {
           inlineValueUnit && inlineValueUnit,
           inlineLabel && d.datum[inlineLabel]
         ].join(' ')
-        d.inlineLabelTextWidth = labelGauger(d.inlineLabelText)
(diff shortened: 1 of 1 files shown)


src/components/Chart/Bars.js
```diff
@@ -333,7 +333,7 @@ const BarChart = props => {
           inlineValueUnit && inlineValueUnit,
           inlineLabel && d.datum[inlineLabel]
         ].join(' ')
-        d.inlineLabelTextWidth = labelGauger(d.inlineLabelText)
+        d.inlineLabelTextWidth = labelGauger(d.inlineLabelText || '')
```
tpreusse, line 336: Darn didn't notice that. But IMO we shouldn't even call labelGauger when we have nothing:
```suggestion
        d.inlineLabelTextWidth = d.inlineLabelText
          ? labelGauger(d.inlineLabelText)
          : 0
```

ovbm: sure, I just used an existing pattern in the code XD

PR #384, opened 2021-04-14
Current Minute

tpreusse:
Move `now` out of discussion context and instead into an `useCurrentMinute` hook. The new hook is then directly used around the DOM element that relies on the current time. This greatly reduced the React conciliation effort necessary.

Before it took at least 40ms, often 60 to 90ms on an M1 with Chrome:
Screenshot 2021-04-14 at 21 26 54

Afterwards we're down to 20ms:
Screenshot 2021-04-15 at 01 14 21

Even tested in IE11 and eliminates the worst freezes. Generally I'd expect this to have a greater positiv impact on slower devices.

### DIFF
--- a/src/components/CommentTeaser/CommentTeaser.js
+++ b/src/components/CommentTeaser/CommentTeaser.js
@@ -15,6 +15,7 @@ import {
   CommentBodyFeaturedText
 } from '../CommentBody/web'
 import { IconLink, Context, Header } from '../Discussion/Internal/Comment'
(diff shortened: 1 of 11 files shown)


ovbm (approved):
After looking at it for 15min I understand what it does. 🥳

tpreusse:
Thx for the review. It is quite nitty gritty. Not 100% sure of the optimization of only using one timeout is truly worth the extra complexity but feels wrong to have 100 of timers running on a discussion page.

PR #386, opened 2021-04-19
fix(ShareImage): swap editorial default font

ovbm:
The default text of ShareTafeln is actually the Republik Font (serifTitle) and not the GT America (serifBold). See example from sketch file: 

Screenshot 2021-04-19 at 10 11 29
Screenshot 2021-04-19 at 10 11 45

GT America is never used in ShareImages, which lucky for us means we can just swap the two fonts.

Sketch file for reference:
https://drive.google.com/file/d/1ywqwE-We-izYf1psxtvll1q4fuw5gi3d/view?usp=sharing

### DIFF
--- a/src/components/ShareImage/ShareImagePreview.js
+++ b/src/components/ShareImage/ShareImagePreview.js
@@ -36,7 +36,7 @@ const styles = {
     width: '100%',
     whiteSpace: 'pre-wrap',
     textAlign: 'center',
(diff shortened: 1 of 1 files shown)


tpreusse (approved):
> not the GT America (serifBold)

`serifBold` is Rubis. GT America is used for meta ;-)

ovbm:
Genau! XD

PR #387, opened 2021-04-23
feat(ShareImagePreview): warn users of padding overflow in ShareImage…

ovbm:
https://user-images.githubusercontent.com/20746301/115880722-532ea380-a44b-11eb-99a0-f7cd4f6d4298.mov

(only shows red background in preview mode)

### DIFF
--- a/src/components/ShareImage/ShareImagePreview.js
+++ b/src/components/ShareImage/ShareImagePreview.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useRef, useEffect, useState } from 'react'
 import { css } from 'glamor'
 import { fontStyles } from '../../theme/fonts'
 import { imageStyle } from './SharePreviewTwitter'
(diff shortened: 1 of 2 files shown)


annatraussnig:
Great idea. 
However: the implementation feels very dramatic and I can see people freaking out at the sigh of this red-blood ocean. How about toning it down a little (e.g. showing a dashed border instead of a red background) and adding a description of the problem (to begin with, at least – I imagine this could be removed once everyone is familiar with the tool)?

ovbm:
true, will make an iteration once I get the backends running on my M1

ovbm:
New and improved, less in your face and with a not offering solutions to get rid of the warning:

I reduced the default font size to 56, because 60 would overflow immediately on some images.

ovbm:
@annatraussnig for your consideration :D

annatraussnig (approved):
LGTM – let's just test the actual generated share image (`ARTICLE_URL?extract=share`) before deploying to master

PR #388, opened 2021-04-28
Comment embed

annatraussnig:


src/templates/Article/index.js
```diff
@@ -514,6 +523,35 @@ const createSchema = ({
...
+                      loading={liveData.loading}
+                      render={() => {
+                        return (
+                          <TeaserEmbedComment
+                            nodeData={nodeData}
+                            liveData={liveData}
+                            t={t}
+                            Link={CommentLink}
+                          />
+                        )
```
annatraussnig, line 544: @tpreusse this part still doesn't handle well the case where an id doesn't exist. The error gets returned from the BE, but is subsequently "erased", which causes all sorts of trouble, but i'mnot able to pinpoint where. Happens both for FE and Publikator. Any advice welcome :)

src/templates/Article/index.js
```diff
@@ -95,9 +98,13 @@ const addProgressProps = rule => ({
 export const COVER_TYPE = 'COVERFIGURE'
 export const DYNAMICCOMPONENT_TYPE = 'DYNAMICCOMPONENT'
 
+const LAZYLOADER_COMMENT_HEIGHT = 250
```
tpreusse, line 101: this was just a guess, please adjust to the usual height.

src/components/TeaserEmbedComment/index.js
```diff
@@ -0,0 +1,112 @@
+import React from 'react'
+import * as Comment from '../Discussion/Internal/Comment'
+import { parse } from '@orbiting/remark-preset'
```
tpreusse, line 3: todo: refactor this out of the client – the remark parser is quite heavy for the client

tpreusse (commented):
I'll test tomorrow and try to figure out what is going on

src/components/TeaserEmbedComment/index.js
tpreusse: 184kb -> 239kb
Screenshot 2021-05-12 at 19 55 05

annatraussnig: wow. ok, yeah. @patrickvenetz and I had talked about maybe delegating the parsing to the backend (which it already does for other comments). so that's a potential solution – unless you had something schlauer in mind.

tpreusse: one possible solution: do it in publikator frontend – which already ships with the parser afaik and is less code size sensitive – and save the parsed ast to the node data.

tpreusse (approved):

PR #389, opened 2021-05-04
feat(ShareIcon): Android and ChromeOS Icon support

ovbm:
tested with chrome's custom user agent setting:
Screenshot 2021-05-04 at 11 01 54


src/components/Icons/CustomIcons/ShareIcon.js
```diff
@@ -0,0 +1,24 @@
...
+  }, [])
+  if (useAndroidIcon) {
+    return <MdShare {...props} />
+  } else {
+    return (
+      <CustomIconBase {...props}>
+        <path
+          d='M8.41 9.41L7 8L12 3L17 8L15.59 9.41L13 6.83V16.5H11V6.83L8.41 9.41ZM21 20V11H19V20H5V11H3L3 20C3 21.1 3.9 22 5 22H19C19.5304 22 20.0391 21.7893 20.4142 21.4142C20.7893 21.0391 21 20.5304 21 20Z'
+          fill={fill}
+        />
```
tpreusse, line 20:
```suggestion
let defaultIsGoogleOS = undefined

export const ShareIcon = ({ fill, ...props }) => {
  const [isGoogleOS, setIsGoogleOS] = useState(defaultIsGoogleOS)
  useEffect(() => {
    if (isGoogleOS !== undefined) {
      return // prevent unnecessary set state / re-renders
    }
    // set default value for all subsequent renders, e.g. client side navigation
    defaultIsGoogleOS = !!window.navigator.userAgent.match(/Android|CrOS/)
    // re-render this component with certain value
    setIsGoogleOS(defaultIsGoogleOS)
  }, [isGoogleOS])
  if (isGoogleOS === undefined) {
    // empty placeholder, avoid flickering between icons on Android
    return <CustomIconBase {...props} />
  } else if (isGoogleOS) {
    return <MdShare {...props} />
  } else {
    return (
      <CustomIconBase {...props}>
        <path
          d='M8.41 9.41L7 8L12 3L17 8L15.59 9.41L13 6.83V16.5H11V6.83L8.41 9.41ZM21 20V11H19V20H5V11H3L3 20C3 21.1 3.9 22 5 22H19C19.5304 22 20.0391 21.7893 20.4142 21.4142C20.7893 21.0391 21 20.5304 21 20Z'
          fill={fill}
        />
```

ovbm: hmm, but this intrdouces an empty placeholder for non-Google-OS devices as well, which could just load the icon without waiting to know the user agent...

tpreusse: jup, if we want to solve that while also not creating a cache issue we'd need CSS:
- rm empty placeholder
- add `css({'.android &': { display: none }})` to the (iOS) path element here
- add android className to html element via script in <head> section of `republik-frontend` (kind of like [`ColorScheme/Sync`](https://github.com/orbiting/republik-frontend/blob/master/components/ColorScheme/Sync.js))

ovbm: seems overkill for now... share icons are never above the fold.

tpreusse: agree that it's overkill because it's just on the first render for like 1s. But it is above the fold most of the times in the article action bar IMG_4276

ovbm: but only if there's no header image, no?

tpreusse: jup at least on most mobile screens. Desktop screen may fit it all ;-)

ovbm:
let's merge and test?

tpreusse (approved):
@ovbm sounds good

PR #394, opened 2021-05-14
Bar Fixes

tpreusse:
Changes
- sum `showBarValues`
- improve rendering precision of bar segments
- rm thousand separator for minus four digits (-1000 to -9999)

### sum `showBarValues`

Screenshot 2021-05-14 at 16 41 40

### rendering precision

| before | after |
|-|-|
| Screenshot 2021-05-14 at 16 41 03 | Screenshot 2021-05-14 at 16 41 14 |
| Screenshot 2021-05-14 at 16 42 12 | Screenshot 2021-05-14 at 16 42 26 |

### DIFF
--- a/src/components/Chart/Bars.js
+++ b/src/components/Chart/Bars.js
@@ -262,11 +262,11 @@ const BarChart = props => {
         height: style.height,
         segments: barSegments,
         first,
-        max: barSegments.reduce(
+        sumPositiv: barSegments.reduce(
           (sum, segment) => sum + Math.max(0, segment.value),
           0
         ),
-        min: barSegments.reduce(
+        sumNegative: barSegments.reduce(
           (sum, segment) => sum + Math.min(0, segment.value),
           0
         )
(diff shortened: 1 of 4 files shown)


src/components/Chart/Bars.js
```diff
@@ -602,6 +580,34 @@ const BarChart = props => {
                       .map((segment, i) => (
                         <segment.Component key={`seg${i}`} />
                       ))}
+                    {showBarValues && (
+                      <>
+                        {bar.min && (
```
annatraussnig, line 613: I find `min` and `max` a bit confusing. Since the prop is called `showBarValues` how about `negValue`and `posValue`instead?

annatraussnig (approved):

src/components/Chart/Bars.js
tpreusse: Thanks for the feedback: I iterated on the name and ended up with `sumNegative` and `sumPositive`: https://github.com/orbiting/styleguide/pull/394/commits/6c4120f8e9485004e8801315e091590f1330c6a6

PR #396, opened 2021-05-25
fix(ShareImage): add missing line-height

ovbm:
There was no line-height specified on the div containing the share image text. For ShareImages with GT-America, this lead to the scrollHeight sometimes to being higher than the clientHeight. It would also cut off the lower parts of the text:

before:
Screenshot 2021-05-25 at 14 24 13
Screenshot 2021-05-25 at 14 24 58

by setting a proper line-height (html defaults to 1.15, which is not enough for gt america):
Screenshot 2021-05-25 at 14 24 19
Screenshot 2021-05-25 at 14 25 13

### DIFF
--- a/src/components/ShareImage/ShareImagePreview.js
+++ b/src/components/ShareImage/ShareImagePreview.js
@@ -167,6 +167,7 @@ const ShareImagePreview = ({
             fontSize: fontSize || SHARE_IMAGE_DEFAULTS.fontSize,
             color: inverted ? '#FFF' : '#000',
             width: shareImage && '80%',
+            lineHeight: 1.25,
             border:
               preview && textContainerOverflow ? '1px dotted red' : 'none',
             maxHeight:


src/components/ShareImage/ShareImagePreview.js
```diff
@@ -167,6 +167,7 @@ const ShareImagePreview = ({
             fontSize: fontSize || SHARE_IMAGE_DEFAULTS.fontSize,
             color: inverted ? '#FFF' : '#000',
             width: shareImage && '80%',
+            lineHeight: 1.25,
```
tpreusse, line 170: The current default value in Chrome seems to be 1.15. Could this increase potentially lead to old image being to big?

ovbm: yes. is there a way around? It should have been 1.25 from the beginning. We could limit potential problems by limiting the line height changes to formats that have gt-america, but it would mean that other formats would remain with narrow line-heights...

ovbm: since it hasn't been that long since we introduced the feature, I'd volunteer to go back to all 30 or so articles to check and clean up errros.

ovbm: which we need to do anyways, because there might be some issues with the lowercase characters "g" "j" etc.

tpreusse (approved):

PR #397, opened 2021-05-25
SeriesNav

ovbm:


src/components/TeaserCarousel/Row.js
```diff
@@ -73,12 +73,29 @@ const styles = {
...
+    const target = Array.from(scroller.children)[initialScrollTileIndex]
+    scrollIntoView(target, {
+      time: 400,
+      align: {
+        left: 0,
+        leftOffset: TILE_MARGIN_RIGHT,
+        ...getTop()
+      }
+    })
+  }, [initialScrollTileIndex])
```
tpreusse, line 97: `scrollIntoView` and an animation time is kind of wrong here. What we really want is to pre set `overflow.current.scrollLeft` with any delay or animation and usually while that element ist outside of the view port. While the `getTop` kind of manages the out of viewport part it is hacky and messes with scroll properties of the whole document on initial mount for 400ms.

Currently e.g. the documentation page of the article accidentally automatically gets scrolled by competing commands from grid and row:
Screenshot 2021-06-10 at 13 35 19

Quick fix: `time: 0`. But if you have time please also look into just setting `overflow.current.scrollLeft`.

src/components/SeriesNav/SeriesNav.js
```diff
@@ -0,0 +1,182 @@
...
+          <p {...styles.description}>{series.description}</p>
+        </Center>
+      ) : null}
+
+      <TeaserCarousel
+        grid={!inline}
+        style={{
+          padding: 0,
+          backgroundColor: 'transparent'
+        }}
```
tpreusse, line 129: supporting a style object on a custom component with complex css make it bridle and harder to maintain. Especially since here we're overwriting styles defined within. Could we add a `isSeriesNav` or `noPadding` prop instead and do all the styling inside?

src/components/SeriesNav/SeriesNav.js
```diff
@@ -0,0 +1,182 @@
...
+      >
+        <TeaserCarouselTileContainer
+          initialScrollTileIndex={currentTileIndex}
+          style={{
+            padding: 0
+          }}
+          overflowCentered
```
tpreusse, line 136: See comment above, IMO `isSeriesNav` would be easiest for now

src/components/SeriesNav/SeriesNavTile.js
```diff
@@ -0,0 +1,102 @@
...
+    current || !!PayNote
+      ? {
+          light: {
+            default: '#191919',
+            text: '#f0f0f0'
+          },
+          dark: {
+            default: '#ffffff',
+            text: '#282828'
+          }
```
tpreusse, line 64: import from `theme/colors.js`, would make the inversion more visible too IMO

tpreusse (commented):
First round of feedback.

src/components/TeaserFeed/Lead.js
```diff
@@ -2,15 +2,15 @@ import React from 'react'
...
 import { convertStyleToRem } from '../Typography/utils'
 
 const styles = {
   main: css({
-    ...convertStyleToRem(serifRegular17),
+    ...convertStyleToRem(serifRegular15),
     margin: '0 0 5px 0',
     [mUp]: {
-      ...convertStyleToRem(serifRegular19)
+      ...convertStyleToRem(serifRegular17)
```
tpreusse, line 13: let's not just change font-size

src/components/SeriesNav/SeriesNav.js
```diff
@@ -0,0 +1,179 @@
...
+        <ColorContextLocalExtension localColors={localColors}>
+          {PayNote ? (
+            <PayNote inline={inline} />
+          ) : (
+            <hr
+              {...styles.hline}
+              {...colorScheme.set('borderColor', 'divider')}
+            />
+          )}
+        </ColorContextLocalExtension>
```
tpreusse, line 162: Let's minimise the `ColorContextLocalExtension` use to only when necessary:

```suggestion
        {PayNote ? (
          <ColorContextLocalExtension localColors={localColors}>
            <PayNote inline={inline} />
          </ColorContextLocalExtension>
        ) : (
          <hr
            {...styles.hline}
            {...colorScheme.set('borderColor', 'divider')}
          />
        )}
```

src/components/SeriesNav/SeriesNavTile.js
```diff
@@ -0,0 +1,102 @@
...
+            <SeriesNavTileContent
+              inline={inline}
+              episode={episode}
+              current={current}
+              ActionBar={ActionBar}
+              Link={Link}
+              onEpisodeClick={onEpisodeClick}
+            />
+          )}
+        </ColorContextLocalExtension>
```
tpreusse, line 96: let's minimize ColorContextLocalExtension use, I'd do following above the return:
```
const isInverted = !!(current || PayNote)
const content = PayNote ? (
  <PayNote />
) : (
  <SeriesNavTileContent
    inline={inline}
    episode={episode}
    current={current}
    ActionBar={ActionBar}
    Link={Link}
    onEpisodeClick={onEpisodeClick}
  />
)
```

and then here
```
        {isInverted ? <ColorContextLocalExtension localColors={localColors}>
          {content}
        </ColorContextLocalExtension> : content}
```

tpreusse (commented):
Through with review. One more thing below.

src/components/TeaserFeed/Lead.js
ovbm: lol XD

tpreusse (approved):

PR #398, opened 2021-06-09
Bug fixes and mini devs

annatraussnig:
### Customizable Tooltip 

Works on map (points && features), and scatterplots. Syntax modelled after our translation strings. E.g.:
```
tooltipLabel: '{name}, {Land}',
tooltipBody: 'CO<sub>2</sub> footprint: {value} tons per year per person.\nThinking of moving to {name}?'
```
#### Maps (Points)
Screenshot 2021-06-09 at 14 52 56

#### Maps (Features)
Screenshot 2021-06-09 at 14 54 05

#### Scatter Plots
Screenshot 2021-06-09 at 14 58 46

### Timebars: Columns
_This one cames at the cost of a significant fight against entropy. I tried to do the right thing and refactor both `TimeBar`and `Line`, in order to share most of the column-specific code and not add another layer of complexity in the component itself. I also split the `TimeBar` file and extracted a bunch of functions, single-responsability-ish style. The feature was fairly trivial to implement once this was done._

Screenshot 2021-06-09 at 15 02 00

### Timebars: Ordinal Scale
Screenshot 2021-06-17 at 09 50 43
Screenshot 2021-06-17 at 09 50 51

### Bars & Lollipops: Skip Y Labels Option
Screenshot 2021-06-09 at 15 05 28

### Bars & Lollipops: Show Unit
Screenshot 2021-06-09 at 15 06 38

### Maps: Custom Choropleth Legend
In this example `über 70%` instead of `70% bis 68%`

Screenshot 2021-06-09 at 15 07 27

#### Maps: recognise 0 as a value
Screenshot 2021-06-09 at 16 14 41

#### Lines: fix y axis number formatting
Before: 
Screenshot 2021-06-09 at 15 11 29
After:
Screenshot 2021-06-09 at 14 52 22

#### Lines: fix missing inline legend 
Before:
file

After:
Screenshot 2021-06-09 at 16 10 38


ovbm:
WOW!

annatraussnig:
Yeah, it's a lot 😅 
You can split the review with Thomas...

src/components/Chart/Bars.js
```diff
@@ -747,7 +748,9 @@ export const propTypes = {
   inlineLabelPosition: PropTypes.string,
   tLabel: PropTypes.func.isRequired,
   description: PropTypes.string,
-  showBarValues: PropTypes.bool
+  showBarValues: PropTypes.bool,
+  skipYLabels: PropTypes.bool,
```
tpreusse, line 752: could this be auto computed when `props.color === props.y` and auto turn on color legend?

src/components/Chart/Bars.js
```diff
@@ -747,7 +748,9 @@ export const propTypes = {
...
+  xUnit: PropTypes.string
```
tpreusse, line 753: rename to `xUnit`, only scatterplots can have different values on `x` and `y`

src/components/Chart/Lines.layout.js
```diff
@@ -200,43 +204,47 @@ export default props => {
...
-        }
-      })
+      const linesWithLabels = lines.map(
+        addLabels(color, colorAccessor, labelFilter, yFormat, y, props)
+      )
       let labelsMoved = 0
       labelsMoved += calculateAndMoveLabelY(linesWithLabels, 'start')
       labelsMoved += calculateAndMoveLabelY(linesWithLabels, 'end')
-      yNeedsConnectors = yNeedsConnectors || labelsMoved > 1
+      yNeedsConnectors = yNeedsConnectors || labelsMoved >= 1
```
tpreusse, line 247: this currently leads to connectors also showing up with two lines when labels were just move a tiny bit
Screenshot 2021-06-15 at 12 22 40
(bottom chart «Mittelstandsinitiative» the second label was moved slightly)

IMO we should only do it when we have more than two lines:
```suggestion
      yNeedsConnectors = yNeedsConnectors || (labelsMoved >= 1 && linesWithLabels.length > 2)
```

src/components/Chart/Maps.docs.md
```diff
@@ -774,40 +901,43 @@ JU,0.257
       "choropleth": true,
       "numberFormat": ".0%",
       "thresholds": [0.3,0.4,0.5,0.6,0.7],
+      "colorLegendLabels": ["unter 30%", "", "", "", "", "über 70%"],
```
tpreusse, line 904: not quite happy with that API yet, that an empty string triggers the default label is not intuitive and skipping is not possible, let's discuss if we should automatically do nicer labels for edge values or a more explicit API

src/components/Chart/TimeBars.js
```diff
@@ -584,13 +283,25 @@ export const propTypes = {
...
+  columnSort: sortPropType,
+  columnFilter: PropTypes.arrayOf(
+    PropTypes.shape({
+      title: PropTypes.string.isRequired,
+      test: PropTypes.string.isRequired
+    })
+  ),
+  columns: PropTypes.number.isRequired,
+  minInnerWidth: PropTypes.number.isRequired,
+  skipRowPadding: PropTypes.bool
```
tpreusse, line 297: Todo for me: look into what this does.

src/components/Chart/ScatterPlots.js
```diff
@@ -318,45 +326,46 @@ const ScatterPlot = ({
...
+            size: sizeFormat(value.size)
+          }
+          const contextT = text => {
+            const replacements = getReplacementKeys(text).reduce(
+              (acc, replacementKey) => {
+                acc[replacementKey] =
+                  acc[replacementKey] ||
+                  value[replacementKey] ||
+                  value.datum[replacementKey] ||
+                  props[replacementKey]
```
tpreusse, line 342: what are the uses cases for exposing props? I'd prefer a well defined set of replacement ala:
```
{
  value: 10000,
  formattedValue: '10'000',
  ...value.datum // all data column values
}
```

src/components/Chart/TimeBars.docs.md
```diff
@@ -756,3 +756,111 @@ year,value,color
...
+      "xTicks": [
+        "< 1",
+        "1–3",
+        "3–5",
+        "5–10",
+        "10–20",
+        "20–30",
+        "30–50",
+        "> 50"
+      ]
```
tpreusse, line 799: this overlaps on mobile:
Screenshot 2021-06-15 at 12 44 23

tpreusse (commented):
Awesome work.

Here is a few comments from a first round of prop types & docs review.

src/components/Chart/Maps.docs.md
annatraussnig: yeah I was a bit on the fence regarding that one. my other idea would be to be able to do something like `["unter 30%", "auto", "auto", "auto", "auto", "über 70%"]`
would it sit better with you?

src/components/Chart/ScatterPlots.js
annatraussnig: we also have the `unit` (or `xUnit`and `yUnit`) which could be useful. But yes that's about it.

src/components/Chart/TimeBars.js
annatraussnig: the `skipRowPadding just puts padding between the columns, but not at the end of the rows (which makes sense for timebars, but is different for lines/slopes)

src/components/Chart/Bars.js
tpreusse: meant to write" rename to `unit` (sorry for the non-sensical initial comment)

src/components/Chart/Bars.js
annatraussnig: and when there is more than one set of data (like in the example)

src/components/Chart/TimeBars.js
tpreusse: Ah I see, so more of an internal config? IMO we should not expose it and just hard code `skipRowPadding: true` when calling the util.

src/components/Chart/TimeBars.docs.md
annatraussnig: is this a problem code though or rather an example problem? part of me feels like if one wants to use this type of histogram, one should chose margins/labels/column number than will render ok on mobile too. (Which I clearly didn't do here – granted.)

src/components/Chart/TimeBars.js
annatraussnig: exactly. i just thought, if you got it, flaunt it. 🤣  
but maybe it adds unnecessary noise indeed.

src/components/Chart/TimeBars.docs.md
annatraussnig: rotate 45° on mobile if needed.

src/components/Chart/ScatterPlots.js
annatraussnig: + add replacement info in docs

src/components/Chart/Maps.docs.md
annatraussnig: `thresholdConfig`

```
[
   {"label": "Weniger als 30%"},
   {minValue: 0.3, "label": "30% bis 40%"},
   {minValue: 0.4},
   {minValue: 0.5, "label": "Über 50%"}
]
```

tpreusse (approved):
Awesome work

PR #404, opened 2021-06-24
Scatter annotations

tpreusse:
Screenshot 2021-06-25 at 01 41 59

Not 100% happy with the annotation label text placement, skipped in example for now.


tpreusse:
Integrated into https://github.com/orbiting/styleguide/pull/405 with column support

PR #407, opened 2021-07-02
fix(OverlayToolbar): align close icon right

annatraussnig:
Make OverlayToolbarClose consistent with our actual layout and conventions

### DIFF
--- a/src/components/Overlay/OverlayToolbar.js
+++ b/src/components/Overlay/OverlayToolbar.js
@@ -5,7 +5,7 @@ import { css } from 'glamor'
 import { CloseIcon } from '../Icons'
 import { useColorContext } from '../Colors/useColorContext'
 import { mUp } from '../../theme/mediaQueries'
-import { sansSerifMedium16 } from '../Typography/styles'
+import { sansSerifMedium16, sansSerifRegular15 } from '../Typography/styles'
 
 export const height = 48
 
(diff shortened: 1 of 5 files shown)


tpreusse:
Doc page looks broken now:
Screenshot 2021-07-05 at 14 44 50


Maybe we should also document how to do titles for the most common use case:
Screenshot 2021-07-05 at 14 32 21

A nice API could be:
```js
<OverlayToolbar title='Teilen' onClose={onCloseFn} />
```

`title` being a `PropTypes.node` – also supporting custom elements:
```js
<OverlayToolbar title={<><MyTabButton>Tab 1</MyTabButton> <MyTabButton>Tab 2</MyTabButton></>} onClose={onCloseFn} />
```

annatraussnig:
that's an idea. i've got a bit of time today, i can clean up the overlay api.

annatraussnig:
ready for rereview

src/components/Overlay/OverlayToolbar.js
```diff
@@ -47,41 +52,31 @@ const styles = {
...
+    color: '#fff',
+    pointerEvents: 'none',
+    position: 'absolute',
+    right: 60,
+    top: 9
   })
 }
 
-export const OverlayToolbar = ({ children }) => {
+const Tag = ({ text }) => {
```
tpreusse, line 70: Is this being used? Maybe I'm missing something or an artefact from trying things out?

tpreusse (approved):
Awesome, it's so simple now. Thx

src/components/Overlay/OverlayToolbar.js
annatraussnig: precisely. i remove

PR #410, opened 2021-07-29
feat(CommentComposer): Allows a label to be displayed 

ovbm:
Uses the DiscussionContext, wich has been expanded in the example (and will be expanded in the frontend) to include a `getLabel` function. The function accepts the input value and can include arbitrarily complex logic to return warning labels, which will be displayed at the bottom of the text input. In the example, it warns the user to escape the "*-character" if it detects it in the text input field.

### DIFF
--- a/src/components/Discussion/Composer/CommentComposer.js
+++ b/src/components/Discussion/Composer/CommentComposer.js
@@ -298,11 +298,11 @@ export const CommentComposer = props => {
           <MaxLengthIndicator maxLength={maxLength} length={textLength} />
         )}
       </div>
-      {label ? (
+      {label && (
         <div {...styles.label}>
           <Label>{label}</Label>
         </div>
-      ) : null}
+      )}
 
       <Loader
         loading={preview.loading && !(preview.comment && preview.comment.embed)}
--- a/src/components/Discussion/Composer/CommentComposer.js
+++ b/src/components/Discussion/Composer/CommentComposer.js
@@ -94,7 +94,7 @@ export const CommentComposer = props => {
   /*
    * Get the discussion metadata and action callbacks from the DiscussionContext.
    */
-  const { discussion, actions, getLabel } = React.useContext(DiscussionContext)
+  const { discussion, actions } = React.useContext(DiscussionContext)
   const { id: discussionId, tags, rules, displayAuthor, isBoard } = discussion
   const { maxLength } = rules
 
@@ -191,7 +191,9 @@ export const CommentComposer = props => {
   const onChangeText = ev => {
     const nextText = ev.target.value
     setText(nextText)
-    setLabel(getLabel(nextText))
+    if (actions.getLabel) {
+      setLabel(actions.getLabel(nextText))
+    }
     try {
       localStorage.setItem(localStorageKey, ev.target.value)
     } catch (e) {
--- a/src/components/Discussion/Composer/CommentComposer.js
+++ b/src/components/Discussion/Composer/CommentComposer.js
@@ -12,6 +12,7 @@ import { convertStyleToRem } from '../../Typography/utils'
 import { Embed } from '../Internal/Comment'
 import { useDebounce } from '../../../lib/useDebounce'
 import { useColorContext } from '../../Colors/useColorContext'
+import { Label } from '../../Typography'
 import Loader from '../../Loader'
 
 const styles = {
@@ -41,6 +42,9 @@ const styles = {
     position: 'absolute',
     bottom: 6,
     left: 8
+  }),
+  label: css({
+    padding: '8px 8px 0px 8px'
   }),
   withBorderBottom: css({
     borderBottomWidth: 1,
@@ -80,6 +84,7 @@ export const CommentComposer = props => {
    */
   const root = React.useRef()
   const [textarea, textareaRef] = React.useState(null)
+  const [label, setLabel] = React.useState(false)
   const textRef = React.useRef()
   const [preview, setPreview] = React.useState({
     loading: false,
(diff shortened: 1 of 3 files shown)


src/components/Discussion/Composer/docs.imports.js
```diff
@@ -66,6 +66,12 @@ export const CommentComposerPlayground = () => {
     actions: {
       openDiscussionPreferences: () => Promise.resolve({ ok: true })
     },
+    getLabel: text => {
+      if (text.indexOf('*') > -1 && text.indexOf('\\*') === -1) {
+        return t('styleguide/CommentComposer/formattingHelp')
```
patrickvenetz, line 71: Maybe go for a specific key name, what it is used for, like:

```suggestion
        return t('styleguide/CommentComposer/formatting/asterisk')
```

src/lib/translations.json
```diff
@@ -30,6 +30,10 @@
       "key": "styleguide/CommentComposer/wait",
       "value": "Sie können {time} wieder einen Beitrag schreiben."
     },
+    {
+      "key": "styleguide/CommentComposer/formattingHelp",
+      "value": "Benutzen Sie ein Backslash vor dem Gendersternchen damit es nicht mit kursiv verwechselt wird:\nLeser\\*in"
```
patrickvenetz, line 35: Hinweis funktioniert nicht, weil die Erklärung fehlt, warum das mit kursiv verwechselt werden würde. Habe das Gefühl, das stiftet Verwirrung. (Auf der [Formatier-Seite](https://www.republik.ch/markdown) klappts, weil die Kursivmachung gleich voran steht.)

Vielleicht eher generell: «Formatieren Sie Ihren Beitrag mit Sonderzeichen (Link Formatier-Seite)»

Vielleicht nur Anweisung: «Damit System Gendersternchen erkennt, stellen Sie einen Backslash voran: Leser\\*in»

Vielleicht ganz kurz: «Gendersternchen setzen: \\*»

src/components/Discussion/Composer/docs.imports.js
```diff
@@ -66,6 +66,12 @@ export const CommentComposerPlayground = () => {
     actions: {
       openDiscussionPreferences: () => Promise.resolve({ ok: true })
     },
+    getLabel: text => {
```
patrickvenetz, line 69: First impressions can be tough, but it felt like `getLabel` needs a home, like `actions.formatHelper`, same as `openDiscussionPreferences`.

src/components/Discussion/Composer/CommentComposer.js
```diff
@@ -186,6 +191,7 @@ export const CommentComposer = props => {
   const onChangeText = ev => {
     const nextText = ev.target.value
     setText(nextText)
+    setLabel(getLabel(nextText))
```
patrickvenetz, line 194: `getLabel` might be not defined, right, because context comes "from outside"? Is it reasonable to harden this?

src/lib/translations.json
ovbm: Hab den Hinweis von der /markdown Seite übernommen. Ich find alle Lösungen ok

src/components/Discussion/Composer/CommentComposer.js
ovbm:
```suggestion
        if (getLabel) {
      setLabel(getLabel(nextText))
    }
```

src/components/Discussion/Composer/docs.imports.js
ovbm: also in actions rein?

src/lib/translations.json
ovbm: Hab "Damit System Gendersternchen erkennt, stellen Sie einen Backslash voran: Leser\*in" übernommen

src/components/Discussion/Composer/docs.imports.js
patrickvenetz: jo

src/components/Discussion/Composer/CommentComposer.js
patrickvenetz:
```suggestion
    getLabel && setLabel(getLabel(nextText))
```

src/lib/translations.json
```diff
@@ -30,6 +30,10 @@
       "key": "styleguide/CommentComposer/wait",
       "value": "Sie können {time} wieder einen Beitrag schreiben."
     },
+    {
+      "key": "styleguide/CommentComposer/formatting/asterisk",
+      "value": "Damit System Gendersternchen erkennt, stellen Sie einen Backslash voran: Leser\\*in"
```
panmau, line 35: I feel like this should say `Damit System Sternchen ...` as the star is also used in other contexts for example for footnotes, date of birth, placeholders and multiplication.

patrickvenetz: I believe, attempt here is to support those being confused when * wrapped text in comment composer becomes cursive font instead of a gender astrisk. It is a very specific "helper" to a user here.

That beeing said, I don't understand just yet what you mean, @panmau, with star being used in other contexts.

@ovbm Wording better, but I slipped you a ton of typos. How about «Um Gendersternchen sichtbar zu machen, stellen Sie einen Backslash voran: \\*»

panmau: It is probably more of an edge case but I was thinking about usages of the star like the following:

```
According to the paper "Sample Book"* of Mr X. (* 05.04.1980) the multiplication 5*10 equals 50. 

* https://example.com
```
That's why I think it would be more fitting to use the more generic word, as the check can't know how the asterisk is used.

patrickvenetz: Got it now. We should be able to find out how frequent astricks appear as a gender bender compared to other uses.

However, after some thinking, maybe not reference gender bender at all but help with "formatting". (Basically what you ment @panmau). Therfor I suggest:

«Um * sichtbar zu machen, stellen Sie ein Backslash voran: \\*»

(Similar might come along with underscore, using it to "genderscore": Bauer_in …)

ovbm:
continues in #411

PR #411, opened 2021-07-29
feat(CommentComposer): Show hints while typing

patrickvenetz:
based on https://github.com/orbiting/styleguide/pull/410

Sample predicate function – to find unescaped asteriks – is a bit more subtle now.

Curiosity led then to improve over label approach. Raising that to an array of hints. A series of predicate functions can run while typing and return one ore more hints.

Went down that rabbithole because there are several other hints we might add e.g. if text contains nothing but a link, show a hint: Posting links are best with some context.

Bildschirmfoto 2021-07-29 um 22 22 23

### DIFF
--- a/src/components/Discussion/Composer/docs.imports.js
+++ b/src/components/Discussion/Composer/docs.imports.js
@@ -70,12 +70,10 @@ export const CommentComposerPlayground = () => {
     composerHints: [
       function formattingAsteriks(text) {
         // Math where asterix is within a word (not next to whitespace) "n*n" for example
-        const hasSingleAsterix = !!text.match(/[^\s:]\*[^\s:]/)
+        const hasSingleAsterix = !!text.match(/[^*\s:]\*[^*\s:]/)
         if (hasSingleAsterix) {
           return (
-            <Label>
-              {t('styleguide/CommentComposer/hints/formattingAsteriks')}
-            </Label>
+            <Label>{t('styleguide/CommentComposer/formatting/asterisk')}</Label>
           )
         }
         return false
--- a/src/components/Discussion/Composer/docs.imports.js
+++ b/src/components/Discussion/Composer/docs.imports.js
@@ -4,10 +4,10 @@ import { createFormatter } from '../../../lib/translate'
 import { Editorial } from '../../Typography'
 
 import { default as Button } from '../../Button'
-import { SecondaryAction } from '../Internal/Composer'
 import { CommentComposer } from './CommentComposer'
 import { DiscussionContext } from '../DiscussionContext'
-import { MarkdownIcon, MoodIcon, StarsIcon } from '../../Icons'
+import { MarkdownIcon, EtiquetteIcon, StarsIcon } from '../../Icons'
+import IconButton from '../../IconButton'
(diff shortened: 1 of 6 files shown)


ovbm:
love, 
only thing: I think the styling of the hint area should be done in the Composer component and only minimally in the components the hint function returns.

patrickvenetz:
Oh, we're at odds then.

Not styling hints in composer itself – apart from a wrapping div container – allows for remarkable flexibility to those providing a discussion context.

I understand a hint much broader than just a string. Could be links, fetching helpful information, show some interactive elements, heck, even GIFs.

ovbm:
I'm not saying hints should be strings. I'm saying there should be a base level consistency in the context/container they are rendered in. That could be just padding, or it could be a uniform background.

ovbm:
Current minimal implementation for Gender* only:
https://user-images.githubusercontent.com/20746301/129751847-1f1e8885-1d51-4412-84fe-2cdf2764834c.mov

src/components/Discussion/Internal/Composer/Actions.js
```diff
@@ -98,11 +77,7 @@ export const Actions = ({
   }, [colorScheme])
   return (
     <div {...styles.root}>
-      {composerSecondaryActions && (
-        <div {...styles.secondaryActions} {...styleRules.secondaryAction}>
-          {composerSecondaryActions}
-        </div>
-      )}
+      {composerSecondaryActions && composerSecondaryActions}
```
tpreusse, line 80:
```suggestion
      {composerSecondaryActions}
```

tpreusse (approved):
LGTM

src/components/Discussion/Composer/docs.imports.js
```diff
@@ -66,15 +67,23 @@ export const CommentComposerPlayground = () => {
     actions: {
       openDiscussionPreferences: () => Promise.resolve({ ok: true })
     },
+    composerHints: [
+      function formattingAsteriks(text) {
+        // Math where asterix is within a word (not next to whitespace) "n*n" for example
+        const hasUnescapedAsterix = !!text.match(/[^\\*\s:]\*[^*\s:]/)
+        if (hasUnescapedAsterix) {
+          return (
+            <Label>{t('styleguide/CommentComposer/formatting/asterisk')}</Label>
```
panmau, line 76: The * is called asterisk. The uses with the typo `asteriks` and `asterix` should probably be fixed.

panmau, line 76:
```suggestion
      function formattingAsterisk(text) {
        // Math where asterisk is within a word (not next to whitespace) "n*n" for example
        const hasUnescapedAsterisk = !!text.match(/[^\\*\s:]\*[^*\s:]/)
        if (hasUnescapedAsterisk) {
          return (
            <Label>{t('styleguide/CommentComposer/formatting/asterisk')}</Label>
```

ovbm: thx XD

patrickvenetz:
Wait, do I ship that? @ovbm

ovbm:
i'll do it

PR #418, opened 2021-09-07
Add xUnit to Time Bar Chart

felixmichel:
Add the option to add an xUnit to a Time Bar. In the config file, one can add an xUnit and then the unit will rendered at the last available bar.
Bildschirmfoto 2021-09-07 um 11 57 49

### DIFF
--- a/src/components/Chart/Bars.js
+++ b/src/components/Chart/Bars.js
@@ -20,7 +20,7 @@ import {
   unsafeDatumFn,
   subsup,
   getTextColor,
-  getLastItemFromArray
+  isLastItem
 } from './utils'
 import ColorLegend from './ColorLegend'
 import { createTextGauger } from '../../lib/textGauger'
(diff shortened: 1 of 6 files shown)


src/components/Chart/Bars.js
```diff
@@ -19,7 +19,8 @@ import {
   deduplicate,
   unsafeDatumFn,
   subsup,
-  getTextColor
+  getTextColor,
+  getLastItemFromArray
```
annatraussnig, line 23: `getLastItemFromArray`is a bit of a ~mouthful~ readful (plus it's a little misleading, you don't actually get anything back). I'd stick to `last` or `isLast`/ `isLastItem` is `last`is a little too terse for your taste.

src/components/Chart/XAxis.js
```diff
@@ -24,7 +25,9 @@ const XAxis = ({
   x,
   xDomain,
   format,
-  strong
+  strong,
+  xUnit,
+  yScaleChangeDirection
```
annatraussnig, line 30: still the old property name here ;)

annatraussnig (changes requested):

src/components/Chart/XAxis.js
felixmichel: fixed

annatraussnig (approved):
Rock'n'roll!

PR #419, opened 2021-09-08
fix(leftLabel): add boolean to align annotations on left side

felixmichel:
Add options for some text alignment and text labels on the left side

- Using two points (x1 and x2) the boolean leftLabel removes the x1 circle and draws a line from the YAxis to the x2 point
- The text alignment of X annotations can be modified by using textAlignment with the options left, right and center, if undefined it defaults to the previous system.

Bildschirmfoto 2021-09-08 um 12 25 44

### DIFF
--- a/src/components/Chart/TimeBars.docs.md
+++ b/src/components/Chart/TimeBars.docs.md
@@ -992,27 +992,28 @@ age,date,value
       "label": "Schätzung",
       "showValue": false,
       "position": "bottom",
-      textAlignment: 'right',
+      textAlignment: "right",
     },
     {
-      "x": '74',
+      "x": "704",
       "value": 90712,
       "label": "Tatsächlich",
       "showValue": false,
       "position": "bottom",
-      "textAlignment": 'right'
+      "textAlignment": "right"
 
     },
     {
-      x1: '0',
-      x2: '70',
-      value: 68139,
-      label: '70. Perzentile',
-      showValue: true,
-      unit: 'Franken',
-      position: 'bottom',
-      ghost: true,
-      leftLabel: true
+      "x1":" "0",
+      "x2": "70",
+      "value": 68139,
+      "label": "70. Perzentile",
+      "showValue": true,
+      "unit": "Franken",
+      "position": "bottom",
+      "ghost": true,
+      "textAlignment": "left",
+      "leftLabels": true
     },
   ]
 }}
--- a/src/components/Chart/TimeBars.docs.md
+++ b/src/components/Chart/TimeBars.docs.md
@@ -991,15 +991,30 @@ age,date,value
       "value": 6000,
       "label": "Schätzung",
       "showValue": false,
-      "position": "bottom"
+      "position": "bottom",
+      textAlignment: 'right',
     },
     {
-      "x": 74,
+      "x": '74',
       "value": 90712,
       "label": "Tatsächlich",
       "showValue": false,
-      "position": "bottom"
-    }
+      "position": "bottom",
+      "textAlignment": 'right'
+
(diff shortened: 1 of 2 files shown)


src/components/Chart/TimeBars.docs.md
```diff
@@ -991,15 +991,30 @@ age,date,value
...
+    {
+      x1: '0',
+      x2: '70',
+      value: 68139,
+      label: '70. Perzentile',
+      showValue: true,
+      unit: 'Franken',
+      position: 'bottom',
+      ghost: true,
+      textAlignment: 'left',
```
annatraussnig, line 1015: Didn't mention it before on the call, but don't you think the `textAlignment`should automatically be `left`when `leftLabel`is `true`? The we can have only `leftLabel` option on here.

src/components/Chart/TimeBarsAnnotations.js
```diff
@@ -16,6 +16,19 @@ const labelGauger = createTextGauger(LABEL_FONT, {
...
+      return 'start'
+    case 'center':
+      return 'middle'
+    case 'right':
+      return 'end'
+    default:
+      return 'middle'
+  }
+}
+
```
annatraussnig, line 31: that's a matter on taste but I tend to prefer dictionaries for this (more concise):
```
const textAlignmentDict = {
  left: 'start',
  center: 'middle',
  right: 'end'
}
```

and then 
```
textAnchor = textAlignmentDict[annotation.textAlignment] || 'middle'
```

src/components/Chart/TimeBarsAnnotations.js
```diff
@@ -107,21 +121,30 @@ export const XAnnotation = ({
...
-  let tx = x1 + (x2 - x1) / 2
-  if (x1 + (x2 - x1) / 2 + textSize / 2 > width) {
-    textAnchor = 'end'
-    tx = x2
-    if ((isRange ? x2 : x1) - textSize < 0) {
-      textAnchor = 'start'
-      tx = x1
+  let tx = annotation.leftLabel ? 0 : x1 + (x2 - x1) / 2
+  if (annotation.textAlignment === undefined) {
+    if (x1 + (x2 - x1) / 2 + textSize / 2 > width) {
```
annatraussnig, line 136: i'd put add an else clause to this `if`
```
else {
  textAnchor = 'middle'
}
```
here.
Then you cover all the cases where textAlignment is undefined here and can ignore the undefined case altogether in l. 145

annatraussnig (approved):
3 suggestions for you to consider but generally looking good.

src/components/Chart/TimeBars.docs.md
```diff
@@ -991,15 +991,29 @@ age,date,value
       "value": 6000,
       "label": "Schätzung",
       "showValue": false,
-      "position": "bottom"
+      "position": "bottom",
+      textAlignment: 'right',
```
annatraussnig, line 995: i know this is super annoying with markdown files, but some things having double-quotes and some not 😱

src/components/Chart/TimeBars.docs.md
```diff
@@ -991,15 +991,29 @@ age,date,value
...
+      x1: '0',
+      x2: '70',
+      value: 68139,
+      label: '70. Perzentile',
+      showValue: true,
+      unit: 'Franken',
+      position: 'bottom',
+      ghost: true,
+      leftLabel: true
+    },
```
annatraussnig, line 1016: also the quotes

annatraussnig (approved):
this `leftLabel`feature is really nice ❤️

PR #420, opened 2021-09-09
fix(leftLabel): get the if else clauses right

felixmichel:
An error in the markdown and if else for leftLabel was false.

Time bars is currently not loading correctly, in time bars it should look like this instead:

Bildschirmfoto 2021-09-09 um 11 35 01


patrickvenetz (approved):
LGTM

PR #424, opened 2021-09-14
fix(d3localeFormat): add custom format for percent and minus

felixmichel:
- use small white space between number and percent
- use minus sign

Bildschirmfoto 2021-09-14 um 15 21 38


annatraussnig (approved):
I still can't get over this alcohol example 😆 
Ship it!

annatraussnig (changes requested):
Actually we need to fix the tests `npm run test`

annatraussnig (approved):
🎉

Issue #426, opened 2021-09-15
Blockquote paragraphs lack a color definition


trm217:
# Description
Paragraphs inside a block-quote lack a color assignment for dark-mode. So far the text relied on inheriting the right text-color from the html/body-elements. This caused the unexpected behaviour mentioned in the following [issue on the publikator-frontend package](https://github.com/orbiting/publikator-frontend/issues/296). The unexpected behaviour is showcased in the screenshot below.

## Screenshot
![Faulty styling in preview iFrame](https://user-images.githubusercontent.com/331800/131125160-0e91fb03-5b93-4420-b996-bab32f36e626.png)

PR #427, opened 2021-09-15
fix(BlockQuoteParagraph): no longer rely on inherited text-color

trm217:
## Description
The color of the blockquote-paragraph is now defined on the component level, so that the text-color is no longer based on the inherited style.

## Screenshot
Screenshot 2021-09-15 at 17 39 58


trm217:
Closes #426

annatraussnig (approved):

PR #428, opened 2021-09-17
fix(EditorialNewsletter): Use 3rd wave logo

patrickvenetz:
Bildschirmfoto 2021-09-17 um 15 27 53

(uses assets provided in https://github.com/orbiting/republik-frontend/pull/635)


tpreusse (approved):
LGTM

PR #430, opened 2021-09-21
Gallery fixes

annatraussnig:
- export gallery min width requirement 
- bring group caption up so they match with regular captions

_before_
Screenshot 2021-09-21 at 16 26 06

_after_
Screenshot 2021-09-21 at 16 25 48

Note: while it would have technically been possible to avoid using negative margins, this would require the more convoluted process of getting all the images in a group, checking which ones are in the last row and removing the bottom margin for those.

### DIFF
--- a/src/templates/Article/base.js
+++ b/src/templates/Article/base.js
@@ -21,7 +21,8 @@ import {
   matchHeading,
   matchType,
   matchZone,
-  matchParagraph
+  matchParagraph,
+  imageSizeInfo
 } from 'mdast-react-render/lib/utils'
 
 import {
@@ -33,6 +34,7 @@ import {
   extractImages,
   matchImagesParagraph
 } from './utils'
+import { MIN_GALLERY_IMG_WIDTH } from '../../components/Figure/Image'
 
 const createBase = ({ metaBody, metaHeadlines }) => {
   const link = {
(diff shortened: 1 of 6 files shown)


src/templates/Article/base.js
```diff
@@ -231,6 +235,9 @@ const createBase = ({ metaBody, metaHeadlines }) => {
   const figureCaption = {
     matchMdast: matchParagraph,
     component: FigureCaption,
+    props: (node, index, parent, { ancestors }) => ({
+      groupCaption: parent.identifier === 'FIGUREGROUP'
```
tpreusse, line 239: Wouldn't the parent of the caption be the FIGURE and only the parent of the parent FIGUREGROUP?

annatraussnig: not if it's a group caption

annatraussnig: `FIGURE`s in `FIGUREGROUP`s can have their own `FIGURECAPTION`s, but there are also shared `FIGURECAPTION`s, which are a direct descendant of `FIGUREGROUP`s and are the ones we are targetting here.

tpreusse: make sense, sorry didn't get it earlier

tpreusse (approved):

PR #431, opened 2021-09-23
fix(hasValues): convert numbers to string to evaluate length

felixmichel:
Currently, if we use the chart components in a dynamic component, the value column of the data has to be a string. With this fix, we can also use number data without converting it to strings.


tpreusse (approved):
🎉

PR #432, opened 2021-09-24
Improve method to persist draft-values for input fields inside discussions

trm217:
## Description

This pull-request aims to solve an issue caused by a shared draft for all inputs on a discussion. Now each input is saved on it's own and can thus be saved without overlap.

### Before

https://user-images.githubusercontent.com/30313631/134655241-377cd854-73fc-47b2-a798-78800b9a08cc.mov

### After

https://user-images.githubusercontent.com/30313631/134655493-5e14b998-908e-45f7-ac82-9ba6b660bc9d.mov

(at 6 seconds you see how previous drafts get migrated)

### DIFF
--- a/src/components/Discussion/Composer/CommentComposer.js
+++ b/src/components/Discussion/Composer/CommentComposer.js
@@ -110,7 +110,6 @@ export const CommentComposer = props => {
       return props.initialText
     } else if (typeof localStorage !== 'undefined') {
       try {
-        //return localStorage.getItem(localStorageKey) || ''
         return readDraft(discussionId, commentId) ?? ''
       } catch (e) {
         return ''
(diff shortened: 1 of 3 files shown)


trm217:
fixed #362

trm217:
closes #375 
@panmau thanks a lot for your effort 👍🏻

tpreusse (commented):
Awesome. Generally looks good to me, a few comments to simplify the code a bit.

One thing that is missing in the «after» video: a list of all drafts around the root compose and or automatically opening the reply compose when it has a draft. But I assume that we'll be part of a future frontend PR? To avoid an additional PR you probably want to expose `loadStoredDraftsObject` in `src/lib.js` (currently `commentComposerStorageKey` is exposed).

tpreusse (approved):

panmau:
@trm217 Thanks for getting this over the finish line! :tada:

PR #433, opened 2021-09-29
First steps for a chart context

felixmichel:
- created a chart context for TimeBars, further abstraction will be needed for other chart tyypes

### DIFF
--- a/src/components/Chart/ChartContext.js
+++ b/src/components/Chart/ChartContext.js
@@ -1,0 +1,195 @@
+import React, { createContext, useMemo } from 'react'
+
+import { ascending } from 'd3-array'
+import { scaleLinear, scaleBand } from 'd3-scale'
+import { timeFormat, timeParse } from '../../lib/timeFormat'
+
+import {
+  deduplicate,
+  hasValues,
+  identityFn,
+  xAccessor,
+  getDataFilter,
+  groupInColumns,
+  getColumnLayout
+} from './utils'
+
+import {
+  insertXDomainGaps,
+  normalizeData,
+  getAnnotationsXValues,
+  processSegments,
+  getMax,
+  getMin
+} from './TimeBars.utils'
+
+// TODO: use padding instead fo axis bottom height
+const COLUMN_TITLE_HEIGHT = 24
+const AXIS_BOTTOM_HEIGHT = 24
+const PADDING_TOP = 24
+const COLUMN_PADDING = 20
+const PADDING_SIDES = 20
+
+export const ChartContext = createContext()
+
+export const ChartContextProvider = props => {
+  const {
+    values,
+    width,
+    mini,
+    xAnnotations,
+    xScale,
+    xInterval,
+    xIntervalStep,
+    domain,
+    padding,
+    yScaleInvert,
+    height: innerHeight
+  } = props
+
+  let xParser = identityFn
+  let xParserFormat = identityFn
+  let xNormalizer = identityFn
+  let xFormat = identityFn
+  let xSort = (a, b) => 0
+  if (xScale === 'time') {
+    xParser = timeParse(props.timeParse)
+    xParserFormat = timeFormat(props.timeParse)
+    xNormalizer = d => xParserFormat(xParser(d))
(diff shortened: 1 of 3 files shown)


tpreusse (commented):
I found a 3 bugs through visual testing

### TimeBar

x axis format / default has changed for time bar small multiples

Screenshot 2021-10-06 at 20 48 17


### Lines

Value / label switches boldness:

Screenshot 2021-10-06 at 20 49 41

---

X axis is missing / rendered in the wrong place with small multiples:

Screenshot 2021-10-06 at 20 51 00
Screenshot 2021-10-06 at 20 52 11

src/components/Chart/ChartContext.js
```diff
@@ -0,0 +1,215 @@
...
+  height: PropTypes.number.isRequired,
+  innerWidth: PropTypes.number.isRequired,
+  x: PropTypes.func.isRequired,
+  xAxis: axisPropType.isRequired,
+  xDomain: PropTypes.array.isRequired,
+  y: PropTypes.func.isRequired,
+  // inconsistent
+  colorValues: PropTypes.array,
+  colorLegend: PropTypes.bool,
+  colorLegendValues: PropTypes.array,
```
tpreusse, line 192: I re-ordered a bit here, I feel like color stuff should be cleaned up and exclusively use via context.

E.g. the context could just provide a `color(datum) => cssColorValue` function (accessor and mapper in one) and a `colorLegendValues` (array of `{color: string, label: string}`) and empty if disabled via props (`colorLegend = false`) or unnecessary (e.g. only one color, or inline line labels).

```suggestion
  color: PropTypes.func.isRequired,
  colorLegendValues: PropTypes. arrayOf(PropTypes.shape({
    color: PropTypes.string.isRequired,
    label: PropTypes.string.isRequired
  })).isRequired,
```
```

File: src/components/Chart/ChartContext.js [hunk 2]
```diff
@@ -0,0 +1,215 @@
...
+  // only used by timebar
+  colorAccessor: PropTypes.func,
+  xNormalizer: PropTypes.func,
+  // timebar only
+  xValues: PropTypes.array,
+  // line only
+  paddingLeft: PropTypes.number,
+  paddingRight: PropTypes.number,
+  columnHeight: PropTypes.number,
+  yAxis: axisPropType,
```
tpreusse, line 202: this should probably eventually become `isRequired` and also be provided by the time bar context. but can wait for another PR

src/components/Chart/XAxis.js
```diff
@@ -2,72 +2,110 @@ import React from 'react'
...
-          strokeDasharray={line.gap ? '2 2' : 'none'}
-        />
-      ))}
-      {ticks.map((tick, i) => (
-        <g
-          key={tick}
-          transform={`translate(${x(tick) + Math.round(x.bandwidth() / 2)}, 0)`}
-        >
+      {type === 'TimeBar' &&
+        baseLines.map((line, i) => (
```
tpreusse, line 47:
```suggestion
      {baseLines &&
        baseLines.map((line, i) => (
```

tpreusse (commented):
It's coming along nicely and the first patterns start to evolve. I couldn't resist and did some clean up and refactoring myself. Now we already have 8 props that are pretty consistent between the two types and highly likely to also be available for scatter plots:
```js
  groupedData: PropTypes.array.isRequired,
  groupPosition: PropTypes.shape({
    y: PropTypes.func.isRequired,
    x: PropTypes.func.isRequired,
    titleHeight: PropTypes.number // timebar only
  }).isRequired,
  height: PropTypes.number.isRequired,
  innerWidth: PropTypes.number.isRequired,
  x: PropTypes.func.isRequired,
  xAxis: axisPropType.isRequired,
  xDomain: PropTypes.array.isRequired,
  y: PropTypes.func.isRequired,
```

I added a few suggestions. Generally I think we should do one more round with the main focus on colour and then merge.

tpreusse (approved):
I APPROVE THIS REFACTOR 🎉  (and will pray when we'll deploy it 🙈)

PR #436, opened 2021-10-04
fix(TeaserEmbedComment): remove whitespace for multiple embeds in a row

annatraussnig:
### Before
Screenshot 2021-10-04 at 11 43 33

### After
Screenshot 2021-10-04 at 11 42 01


ovbm (approved):
It feels weird not doing this with preuso-selectors and css, but It doesn't seem to be possible...

annatraussnig:
I'm so dead into these schemas atm that it didn't even cross my mind. I'll check if it's possible tomorrow morning 😄

annatraussnig:
So with `loaders` and stuff, the className and css would have to be in the template. Which feels kinda weird.

PR #437, opened 2021-10-07
feat(SeriesNav): adds ability to display series description in grid and disable description inline

ovbm:
Also fixes a small bug, where the min-width of the grid including padding would overflow the viewport on small Phone sizes. Reducing the min-width fixes the issue

### DIFF
--- a/src/components/SeriesNav/SeriesNav.js
+++ b/src/components/SeriesNav/SeriesNav.js
@@ -53,12 +53,7 @@ function SeriesNav({
   aboveTheFold,
   seriesDescription
 }) {
-  const showSeriesDescripion =
-    seriesDescription ?? (
-      inline
-      ? true
-      : undefined
-    )
+  const showSeriesDescripion = seriesDescription ?? (inline ? true : undefined)
 
   const currentTile =
     repoId &&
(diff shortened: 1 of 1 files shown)


src/components/SeriesNav/SeriesNav.js
```diff
@@ -110,22 +121,26 @@ function SeriesNav({
...
+                      </Link>
+                    )
+                  })}
+              </InfoBoxText>
+            )}
           </InfoBox>
           {inlineAfterDescription}
         </Center>
+      ) : showSeriesDescripion ? (
+        <Editorial.P {...styles.description}>{series.description}</Editorial.P>
```
tpreusse, line 143: We should avoid setting [css rules on custom elements](https://styleguide.republik.ch/dev/colors#avoid-custom-elements).

```suggestion
        <Editorial.P style={{padding: '0px 15px'}}>{series.description}</Editorial.P>
```

or

```suggestion
        <div {...styles.description}>
          <Editorial.P>{series.description}</Editorial.P>
        </div>
```

It does currently work:
Screenshot 2021-10-08 at 06 03 25

But if any other rule on the element started to set padding it would be random (execution order) dependant which padding would win.

src/components/SeriesNav/SeriesNav.js
```diff
@@ -47,8 +50,16 @@ function SeriesNav({
...
-  aboveTheFold
+  aboveTheFold,
+  seriesDescription
 }) {
+  const showSeriesDescripion =
+    seriesDescription !== undefined
+      ? seriesDescription
+      : inline
+      ? true
+      : undefined
```
tpreusse, line 61: Fancy way of writing with [`Nullish coalescing operator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator):
```suggestion
    seriesDescription ?? (
      inline
      ? true
      : undefined
    )
```

PR #438, opened 2021-10-14
fix(SeriesNav): hide paynote if a user has an active membership

trm217:
## Description
This change is required to hide the PayNote for members when showing statically generated sites, before React has rehydrated the page.

### DIFF
--- a/src/components/SeriesNav/SeriesNavTile.js
+++ b/src/components/SeriesNav/SeriesNavTile.js
@@ -72,6 +72,11 @@ const SeriesNavTile = ({
 
   return (
     <div
+      /* 
+      Hide PayNote for members by setting the hideForMember class. 
+      This class is defined in the MeContext in republik-frontend
+      */
+      className={PayNote ? 'hideIfMember' : undefined}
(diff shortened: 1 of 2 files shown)


src/components/SeriesNav/SeriesNavTile.js
```diff
@@ -72,6 +72,13 @@ const SeriesNavTile = ({
 
   return (
     <div
+      /* 
+      Hide PayNote for members by setting the data-hide-if-member attribute. 
+      This class is defined in the MeContext in republik-frontend
+      */
+      {...(PayNote && {
+        'data-hide-if-member': true
```
tpreusse, line 80: Result will be following HTML
```html
<div data-hide-if-member="true">
```

Same as JSX
```jsx
<div data-hide-if-member>
```

But I saw that you also did the following JSX in the frontend:
```jsx
<div data-hide-if-member=''>
```

Which lead to following *HTML*:
```html
<div data-hide-if-member>
```

I think is true the way to go. But let's make sure we're consistent.

PR #443, opened 2021-10-22
Upgrade bundling, export everything from a single file and add support for TypeScript

trm217:
## Description

Rollup.js has been added as the module-bundler. The code is bundled for both commonjs and es-modules. Additionally TypeScript was added. The output-directory also changed from `/lib` to `/dist`.

## Todo
- [x] Upgrade ESLint to match the eslint configuration of the frontend repository.


rollup.config.js
```diff
@@ -0,0 +1,65 @@
...
+      useTsconfigDeclarationDir: true
+    }),
+    babel({
+      babelHelpers: 'bundled',
+      exclude: 'node_modules/**'
+    }),
+    commonjs(),
+    fileSize()
+  ],
+  external: [
```
tpreusse, line 42: Should we potentially do `Object.keys(require("./package.json").peerDependencies)` here? Let's also review the peer list—IMO a few of them can no longer be peer, e.g. `downshift` and `lodash`.

trm217:
@tpreusse there are still issues regarding the linters. To properly lint TS files, we need ESLint >= 7.0.0. However React-Scripts V3 needs an ESLint version of ^6.6.0. This can be solved by upgrading React-Scripts to ^4.0.0, but this in return breaks the docs, since Catalog doesn't work with react-scripts V4 🙄

PR #451, opened 2021-11-10
Tabs

ovbm:
Introduces Tabs with several common responsive layout options (scroll, dropdown).

Would be interested to have a quick discussion on if this implementation is good and usefull before going into hardcore code review. 

Screenshot 2021-11-10 at 14 14 58


ovbm:
Made some tweaks regarding layout and styling while implementing it in the frontend. Quite happy with the results:!
next step would be better loading behavior
https://user-images.githubusercontent.com/20746301/141487393-a90c4db4-8f67-4b83-a97f-f24e537f1b81.mov

tpreusse (commented):
First round re scroller. Have to run to a meeting now :)

tpreusse (commented):
Round two :)

src/components/Scroller/docs.md
```diff
@@ -0,0 +1,64 @@
...
+The `Scroller` Component arranges its children in a flex row and allows for horizontal scrolling if the content size exeeds the container size. It is the basis of the `<Tab type='scroll' />`components.
+
+Supported Props
+
+- `children` (ReactNode, required):
+- `center`? (boolean, false): centers the child nodes when container with exeeds content width.
+- `activeScrollItemIndex`? (number, 0): The index of the active scroll item
+- `hideArrows`? (boolean, false): Hides the arrows
+- `arrowSize`? (number, 28): Arrow Size
+- `breakoutWidth`? (number, 0): If the Scroller is in a container with negative horizontal margins, pass those as breakoutWidth to allow for proper padding in full width layouts
```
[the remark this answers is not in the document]
ovbm: I think the issue is that I accidentally made it seem like breakout is dependent on arrows being present. Which it doesn't really, it just relies on the same state as arrows, because "right" just measures if the scroll box is larger than its container which is what is used to use the breakout values or not. I don't want to limit the usability of this component to articles, which is why I'm reluctant to only support a const.

src/lib.ts
```diff
@@ -166,7 +166,8 @@ export { default as SharePreviewTwitter } from './components/ShareImage/SharePre
 
 export { default as TeaserEmbedComment } from './components/TeaserEmbedComment'
 export { SeriesNav } from './components/SeriesNav'
-
+export { default as Tabs } from './components/Tabs'
+export { default as Scroller } from './components/Scroller'
```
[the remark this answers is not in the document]
ovbm: Tabs are just a special case of a Scroller with particular styling of the child nodes. I first hat it so that if you passed an an item with an {element: () } into Tabs, that it would render that instead of the TabItem. But I abandoned that as it got messy with nested buttons...

src/components/Tabs/Scroller.tsx
```diff
@@ -0,0 +1,236 @@
...
+      align: {
+        left: 0,
+        leftOffset: breakoutPadding,
+        ...getTop()
+      }
+    })
+  }, [activeChildIndex])
+
+  useEffect(() => {
+    const scroller = scrollRef.current
```
ovbm, line 106: @tpreusse if i remove the destucturing here, the app throws an error due to unounting of eventlistener of line 129
scroller.removeEventListener('scroll', measure).

ovbm:
@tpreusse new API, much simpler than before:
- no fixed, no dropdown, no specific Tabs component. 
- TabButton with ref, href and onClick
will adapt frontend tabs branch to see if it all works as intended.

src/components/Tabs/Scroller.tsx
```diff
@@ -0,0 +1,236 @@
...
+      setOverflow(scroller.clientWidth < scroller.scrollWidth)
+    }
+    scroller.addEventListener('scroll', measure)
+    window.addEventListener('resize', measure)
+    measure()
+    return () => {
+      scroller.removeEventListener('scroll', measure)
+      window.removeEventListener('resize', measure)
+    }
+  }, [])
```
tpreusse, line 132: Ah I overlooked that the scroll event is listening to the `scroller` itself. I guess moving the assignment outside of the effect and adding `scroller` to the dependency would be the 100% correct solution to ensure that the listener is always attached to the current dom element. But probably the dom element never changes in practice right now. 

```suggestion
  const scroller = scrollRef.current
  useEffect(() => {
    const measure = () => {
      let left = false
      let right = false
      if (scroller.scrollLeft > 0) {
        left = true
      }
      if (scroller.scrollLeft + scroller.clientWidth < scroller.scrollWidth) {
        right = true
      }
      setArrows(current => {
        if (current.left !== left || current.right !== right) {
          return { left, right }
        }
        return current
      })
      setOverflow(scroller.clientWidth < scroller.scrollWidth)
    }
    scroller.addEventListener('scroll', measure)
    window.addEventListener('resize', measure)
    measure()
    return () => {
      scroller.removeEventListener('scroll', measure)
      window.removeEventListener('resize', measure)
    }
  }, [scroller])
```

src/components/Tabs/Scroller.tsx
```diff
@@ -0,0 +1,236 @@
...
+  children,
+  activeChildIndex = 0,
+  center = false,
+  hideArrows = false,
+  arrowSize = 28,
+  breakoutPadding = 0
+}: ScrollerType) => {
+  const scrollRef = useRef<HTMLDivElement>()
+  const [{ left, right }, setArrows] = useState({ left: false, right: false })
+  const [scrollerOverflow, setOverflow] = useState(false)
```
tpreusse, line 88: This can just be deferred from the arrows afais:
```
  const hasOverflow = left || right
```

src/components/Tabs/Scroller.tsx
```diff
@@ -0,0 +1,236 @@
...
+        {...styles.scroller}
+        style={{
+          justifyContent: shouldCenter ? 'center' : 'flex-start'
+        }}
+      >
+        <div
+          style={{
+            flex: shouldCenter
+              ? 1
+              : `0 0 ${scrollerOverflow ? breakoutPadding : 0}px`
```
tpreusse, line 183: why would the padding only be rendered when overflowing? this make the spacing inconsistent between desktop and mobile right now
```suggestion
              : `0 0 ${breakoutPadding}px`
```

src/components/Tabs/Scroller.tsx
```diff
@@ -0,0 +1,236 @@
...
+          }}
+        />
+        {children}
+        <div
+          style={{
+            flex: shouldCenter
+              ? 1
+              : `0 0 ${scrollerOverflow ? breakoutPadding : 0}px`
```
tpreusse, line 191:
```suggestion
              : `0 0 ${breakoutPadding}px`
```

src/components/Tabs/docs.md
```diff
@@ -0,0 +1,227 @@
...
+  >
+    <Link href="/" passHref>
+      <TabButton
+        text="This is a link"
+        isActive={state.activeChildIndex === 0}
+      />
+    </Link>
+  </Scroller>
+</div>
+```
```
tpreusse, line 227: Screenshot 2021-11-17 at 13 45 57


```suggestion
```code|lang-js
<div style={{ margin: '0 -20px' }} >
  <Scroller
    breakoutPadding={20}
    activeChildIndex={0}
    center={true}
  >
    <Link href="/" passHref>
      <TabButton
        text="This is a link"
        isActive
      />
    </Link>
  </Scroller>
</div>
```
```

File: .eslintrc.json [hunk 9]
```diff
@@ -37,7 +37,8 @@
       "parser": "@typescript-eslint/parser",
       "rules": {
         "@typescript-eslint/no-loss-of-precision": "off",
-        "react/prop-types": "off"
+        "react/prop-types": "off",
+        "react/display-name": "off"
```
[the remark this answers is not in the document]
ovbm: it mirrors the rule we already use on line 18 of .eslintrc.json in js files. when it's not off, the pattern React.forwardref(()... throws an eslint error.

src/components/Tabs/Scroller.tsx
ovbm: that was the previous way it was done, which led you to assume that the overflowPadding was dependent on the visibility of the arrows XD. So what's it going to be? This is a bit more declarative and understandable imho :D

tpreusse (approved):

PR #454, opened 2021-11-11
Radio Check

tpreusse:
For the election components we use the checkboxes and radio without labels. I'd like to officially support that use case and optimise the alignment by removing line heights and margins when no label or children are present.

The code also includes the switch from the deprecated `fontFamilies` to `fontStyles`.

Additionally I've moved Radio and Checkbox to it's own docu page:
screencapture-localhost-3003-forms-radio-2021-11-11-15_18_36
screencapture-localhost-3003-forms-checkbox-2021-11-11-15_18_25


src/components/Form/Radio.js
```diff
@@ -1,15 +1,21 @@
...
 
 const styles = {
   label: css({
-    fontSize: 16,
-    lineHeight: '20px',
-    fontFamily: fontFamilies.sansSerifRegular,
     cursor: 'pointer'
   }),
+  withText: css({
+    ...fontStyles.sansSerifRegular,
```
ovbm, line 12: Whats the benefit of using this with pxToRem instead of just ...fontStyles.sansSerifRegular16 ?

ovbm (approved):
LGTM

PR #455, opened 2021-11-16
fix text-decoration in underline style mixin

patte:
Just a tiny thing I discovered:
`text-decoration-skip: ink;` is an invalid property name. You maybe wanted:
`text-decoration-skip-ink: auto;`
No visual change expected, as `auto` seems to be the default.

https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip-ink

### DIFF
--- a/src/lib/styleMixins.js
+++ b/src/lib/styleMixins.js
@@ -6,5 +6,5 @@ export const ellipsize = {
 
 export const underline = {
   textDecoration: 'underline',
-  textDecorationSkip: 'ink'
+  textDecorationSkipInk: 'auto'
 }


tpreusse:
Good catch, this is a relict from 2017 when Chrome supported/needed `text-decoration-skip: ink;` for nicer underlines.

Screenshot 2021-11-16 at 20 05 26

IMO since this is now the default behaviour and pretty much no one uses a 2017 Chrome anymore we can just drop the ink definitions and just use `textDecoration: 'underline'` alone.

tpreusse:
One a second check: Safari 15 still needs it.
Screenshot 2021-11-16 at 20 14 57
Screenshot 2021-11-16 at 20 14 52

PR #458, opened 2021-11-22
Charts Quick Wins

tpreusse:
Changes
- allow `yLines` and `xLines` for axises in line charts
- `yLines` and `xLines` specific to one `column`, same API as `annotations`
- show `unit` in bars and lollipops with `showBarValues`
- line chart docs: rm NA values leading to react warning, NA is not a supported value

Screenshot 2021-11-22 at 23 29 21
Screenshot 2021-11-22 at 23 42 05


felixmichel (commented):
🥇

PR #466, opened 2021-12-20
feat(Button): extends buttons with small and naked styles

ovbm:
Our frontend uses a lot of pseudo-buttons and weird hybrid styled buttons, link as buttons etc... Mostly because our button component is huge and thus not usable in hybrid UI where for example an edit function is not supposed to be the most important UI element on screen.

A collection of our lovely button chrüsimüsi:
Screenshot 2021-12-20 at 16 42 00
Screenshot 2021-12-20 at 16 41 53
Screenshot 2021-12-20 at 16 39 00

This PR proposes two new props to make our styleguide buttons more usable and our frontend less chaotic. A small prop, which renders a small button and a naked prop, which renders a button without backgroundcolor. Together these two props should allow us to use the styleguide button component in 95% of our button needs in the frontend.

naked allows for more drastic differentiation between important actions (like BUY) and less important ones (like "cancel")
Screenshot 2021-12-20 at 17 20 33

small (naked) buttons can be used in our Dialog, and on our Account page for "edit" buttons
Screenshot 2021-12-20 at 17 20 25


tpreusse (commented):
I misunderstood this PR and initially thought it would change much more. I'm fine with introducing the `naked` and `small` style.

Mandatory refactors:
- get ride of `!important`
- re-apply to new ts button file in master

Optional things that could be improved:
- rm `big` style – IMO this was never used and we have custom height if every necessary
- change `hover` color of regular button to not look like `primary` buttons. Even just using `primaryHover` for regular button hover instead of `primary` color as hover would already be great.

PR #469, opened 2022-01-06
Refactor discussion-components

trm217:


src/components/Discussion/Internal/Comment/CommentEmbed.js
```diff
@@ -85,20 +86,23 @@ const dateFormat = timeFormat('%d.%m.%Y %H:%M')
...
 
-export const Embed = ({ comment }) => {
+const propTypes = {
+  embed: PropTypes.object.isRequired,
+  mentioningDocument: PropTypes.object
+}
+
+export const CommentEmbed = ({ embed, mentioningDocument }) => {
   const [colorScheme] = useColorContext()
-  if (!comment || !comment.embed) return null
```
tpreusse, line 100: this file now has crazy amounts of defensive coding, but lake simple checks like this. let's fix this the next time

src/components/Discussion/Tree/BoardComment.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { css } from 'glamor'
+import CommentNode, { CommentProps } from './CommentNode'
+import { CommentEmbed } from '../Internal/Comment'
+import { mUp } from '../../../theme/mediaQueries'
+
+const styles = {
```
tpreusse, line 7: the grid layout here was whack because it created a new grid for every comment and did not align across multiple comments.

src/components/Discussion/Internal/Comment/CommentActions.tsx
```diff
@@ -26,30 +29,54 @@ const styles = {
...
+  const isDesktop = useMediaQuery(mUp)
 
   const now = useCurrentMinute()
   const [colorScheme] = useColorContext()
 
-  const replyBlockedMessage = useMemo(() => {
+  const replyBlockedMessage: string | null = useMemo(() => {
     const waitUntilDate = userWaitUntil && new Date(userWaitUntil)
-    if (waitUntilDate && waitUntilDate > now) {
+    if (waitUntilDate && waitUntilDate.getTime() > now) {
```
tpreusse, line 66: typescript lead to a wrong solution initial here:
https://github.com/orbiting/styleguide/pull/469/commits/5838406169858406e6baf22979e9cea05e372b9a

typecasting was fine:
Screenshot 2022-01-25 at 17 36 06

(of course this also has to do with bad naming in the JS dates api but still typescript sent you in a wrong direction)

tpreusse (approved):
This was quite the pain to get through :D

We definitely need a better typescript strategy—too many `ts-ignore` and even a bug was introduced because of a type warning. Current opinion: converting stuff to TS is good but should be done without changing anything else to be reviewable.

tpreusse:
@trm217 don't worry about it—the code had some issue and it's good that we modernised it a bit.

I think I made it through now 🙈  I shall release tomorrow 🤞

PR #470, opened 2022-01-10
Ob/type forms, translate, reomve colorcontext type error

ovbm:
Types most of our Forms. With the exception of FieldSet and Autocomplete. These two are quite complex and have shown to work as is when using them in frontend .tsx files. 

Testing (see if typescript throws errors) in frontend
[x] Radio 
[x] Dropdown
[x] Slider
[x] Checkbox
[x] FieldSet   
[x] Field
[x] getCSSColor


src/components/Form/Checkbox.tsx
```diff
@@ -12,6 +12,17 @@ const Checkbox = ({
   onChange,
   black,
   error
+}: {
+  children: React.ReactNode | string
```
annatraussnig, line 16: since we use the typed children, any reason not to use `React.FC`?

src/components/Form/DropdownLabel.tsx
```diff
@@ -71,7 +71,24 @@ export const Label = ({
...
+}: {
+  top?: boolean
+  focus?: boolean
+  error?: boolean
+  text?: string
+  children: React.ReactNode | React.ReactNode[]
+  Element?: string
+  field?: boolean
+  value?: string
+  style?: { [key: string]: string }
```
annatraussnig, line 88: I believe this is a good use case for `Record<string, string>`

annatraussnig (commented):
Two small questions

src/components/Form/Checkbox.tsx
ovbm: What would that do?

annatraussnig: basically you get the children for free. (example from the [styleguide](https://styleguide.republik.ch/dev/typescript#reactfc))
```typescript
const ArrowUp: React.FC<{
  size: number
  fill?: string 
}> = ({ size, fill, children }) => (
  <svg fill={fill} width={size} height={size} viewBox='0 0 24 24'>
    <path d='M7.41 15.41L12 10.83l4.59 4.58L18 14l-6-6-6 6z' />
    <path d='M0 0h24v24H0z' fill='none' />
    {children}
  </svg>
)
```

src/components/Form/DropdownLabel.tsx
ovbm: Is it just me or is `{ [key: string]: string }` just infinitely more readable than `Record<string, string>`? Like, if I know nothing of Typescript, I still probably get the former, the latter... i don't know what it truly means :D

src/components/Form/Checkbox.tsx
ovbm: I don't like it. Too much magic...
But I just found out, that "ReactNode allows multiple elements, strings, numbers, fragments, portals, …", so I can omit the `| string` part

annatraussnig: yes I believe you can also get rid of the array.

src/components/Form/DropdownLabel.tsx
ovbm: but done, because why not ^^

src/components/Form/Checkbox.tsx
annatraussnig: We use plenty of magic in other parts of our code (graphql, etc.). But I see your concerns. What I really don't like though, is inconsistency. We can add types one way or another, but if we sometimes use generics, sometimes plain types, it kinda gets messy. Unfortunately, I can't remember reaching a consensus on that, as a team.
Can we have a quick chat on the topic tomorrow or next week?

annatraussnig (approved):
Approved. :)  
Everything looks fine, but we really need to decide how we want to do stuff regarding types. (@tpreusse, @trm217)

src/components/Colors/ColorContext.tsx
```diff
@@ -207,7 +215,9 @@ export const ColorContextProvider = ({
         window.CSS &&
         window.CSS.supports &&
         window.CSS.supports('color', 'var(--color-test)')
-    } catch (e) {}
+    } catch (e) {
+      return
```
tpreusse, line 219: this return will introduce a bug, why did you add it?

ovbm: because my linter complained, I couldn't commit with this in place

ovbm: Screenshot 2022-01-11 at 11 15 11

ovbm: I can either do

`    } catch (e) {
        return ''
      }`

or 

`catch (e) {
      console.error(e)
      return true
    }`

ovbm: (as seen in other places in our code)

tpreusse: the correct solution is on the linter page:
https://eslint.org/docs/rules/no-empty

```
} catch (e) {
  // continue regardless of error
}
```

Adding a return, log or any code where there was no code changes the behaviour. Don't change behaviours because it's done elsewhere, stack overflow says so or the linter complains.

ovbm: thx

tpreusse:
In my opinion follow what the style guide says:
https://styleguide.republik.ch/dev/typescript

(`React.FC` as far as I understand. If not change the style guide?)

Generally using `React.FC` seems good because it defined the return type and regular React props like key – not magic, just type complete – @trm217 what do you think?

ovbm:
ok, will change

trm217:
@tpreusse in general I think FC is a good choice. However I believe that we shouldn't use it if the component doesn't actually accept `children` as a prop. In that case `const Component = ({}: PropsType): ReactElement /* or JSX.Element */` is the better choice in my opinion.

annatraussnig:
@trm217 I like that rule. @ovbm, do you have time to include it somewhere in the Styleguide page as part of this PR, or shall I separately?

ovbm:
@annatraussnig go ahead. I don't understand why using one over the other matters.

src/development/typescript.docs.md
```diff
@@ -35,6 +35,20 @@ const ArrowUp: React.FC<{
 )
 ```
 
+We recommend to use `React.FC` whenever the corresponding component accepts `children` as a prop. If not, better to type a standard Typescript function. Similar example as above, minus children:
+
+```code|lang-js
```
ovbm, line 40: Isn't this typed twice now?
Shouldn't it be:
```
const ArrowUp = ({ size, fill }: { size: number; fill?: string }): ReactNode => (
   <svg fill={fill} width={size} height={size} viewBox='0 0 24 24'>
     <path d='M7.41 15.41L12 10.83l4.59 4.58L18 14l-6-6-6 6z' />
     <path d='M0 0h24v24H0z' fill='none' />
   </svg>
 )
 ```

annatraussnig: 🤦

PR #471, opened 2022-01-11
feat(Buttons): new naked and small styles

ovbm:


tpreusse:
active style is not well defined for naked:
https://user-images.githubusercontent.com/410211/148963621-b01db8d1-8a9b-49e3-9ee7-598bc9c98002.mov

would be nice to add previews of hover and active in the docs for naked like primary and default:
Screenshot 2022-01-11 at 15 42 12

(allows to quickly visually grasp all states and test when modifying, you can use the `simulate='hover'` and `simulate='active'` documentation prop to do so)

ovbm:
thx! missed that
Screenshot 2022-01-11 at 15 53 07

ovbm:
correct. now I did XD

tpreusse (approved):
LGTM, `textSoft` seems a bit light especially on the chequer pattern. But let's see it in live action before optimising.

I moved all the css merge stuff into the memo for slightly better performance on re-renders and detangled the naked logic a bit to also remove calls like this: `colorScheme.getCSSColor(!naked && 'disabled')`, calling `colorScheme.getCSSColor` with false is not supported and will cause a lookup for a color named `'false'`.

ovbm:
fixed a small regression, naked hover wasn't working. will merge now

tpreusse:
@ovbm thx 🙇

PR #472, opened 2022-01-11
fix(Dropdown): fix regression

ovbm:


tpreusse (approved):
LGTM, the other one that would be nice to restore prop forwarding on is slider. E.g. `step` prop is no longer supported there.

PR #476, opened 2022-01-19
fix(Overlay): add types

annatraussnig:


src/components/Overlay/OverlayToolbar.tsx
```diff
@@ -54,7 +54,11 @@ const styles = {
   })
 }
 
-const OverlayToolbarClose = ({ onClick }) => {
+const OverlayToolbarClose = ({
+  onClick
+}: {
+  onClick: () => void
```
trm217, line 60: I'd use `React.MouseEventHandler<HTMLButtonElement>` instead

src/components/Overlay/OverlayToolbar.tsx
```diff
@@ -66,7 +70,10 @@ OverlayToolbarClose.propTypes = {
   onClick: PropTypes.func.isRequired
 }
 
-export const OverlayToolbar = ({ title, onClose, children }) => {
+export const OverlayToolbar: React.FC<{
+  title: string | ReactNode
+  onClose: () => void
```
trm217, line 75: Fine here, since it's abstracted

src/components/Overlay/Overlay.tsx
```diff
@@ -89,16 +104,19 @@ Overlay.propTypes = {
 
 export default Overlay
 
+type MUpStyle = {
+  maxWidth?: number | string
+  marginTop?: number | string
+  marginBottom?: number | string
+}
```
trm217, line 111: Maybe we should type this more generically as React.CSSProperties, since it's spread into the styles object? 🤔

trm217 (approved):
Some suggestions, but LGTM besides that.

src/components/Overlay/Overlay.tsx
annatraussnig: yeah this was to match the `PropTypes` down in the files

src/components/Overlay/OverlayToolbar.tsx
annatraussnig: I am not sure I get what you trying to say. I used `React.FC` since the component makes explicit use of `children` (if that's what you refer to).

PR #477, opened 2022-01-20
Table feature

felixmichel:
add sortable table, current features:

- Sort table by clicking on headers
- Color scale option for specific column
Bildschirmfoto 2022-01-20 um 13 52 20
Bildschirmfoto 2022-01-20 um 13 52 32


src/components/Chart/Table.js
```diff
@@ -0,0 +1,255 @@
...
+    let parsedItem = {}
+    Object.keys(row).forEach(
+      item =>
+        (parsedItem[item] =
+          tableColumns.find(d => d.column === item)?.type === 'number'
+            ? +row[item]
+            : row[item])
+    )
+    parsedData.push({ ...parsedItem })
+  })
```
tpreusse, line 83:
```suggestion
  const numberColumns = tableColumns.filter(d => d.type === 'number').map(d => d.column)
  const parsedData = numberColumns.length
    ? values.map(row => {
      let parsedRow = { ...row }
      numberColumns.forEach(key => {
        if (parsedRow[key] !== undefined) {
          parsedRow[key] = +parsedRow[key]
        }
      })
      return parsedRow
    })
    : [].concat(values)
```

src/components/Chart/Table.js
```diff
@@ -0,0 +1,255 @@
...
+    defaultSortColumn,
+    thresholds,
+    tableColumns
+  } = props
+  const columns = values.columns
+
+  const [sortBy, setSortBy] = useState({
+    key: defaultSortColumn,
+    order: 'desc',
+    enableSorting: !!defaultSortColumn || false
```
tpreusse, line 68:
```suggestion
    order: 'desc'
```
The key is and can be set to `undefined` if it should not sort.

src/components/Chart/Table.js
```diff
@@ -0,0 +1,255 @@
...
+    }
+    if (typeof a[sortBy.key] === 'string') {
+      return sortBy.order === 'desc'
+        ? a[sortBy.key].localeCompare(b[sortBy.key])
+        : b[sortBy.key].localeCompare(a[sortBy.key])
+    }
+    return sortBy.order === 'desc'
+      ? b[sortBy.key] - a[sortBy.key]
+      : a[sortBy.key] - b[sortBy.key]
+  })
```
tpreusse, line 97: `sort` mutates the array – you shouldn't create a new variable because the old one is also changes anyhow.

Also always use the d3-array functions:
```suggestion
  if (sortBy.key) {
    parsedData.sort((a, b) => {
      return sortBy.order === 'desc'
        ? descending(a[sortBy.key], b[sortBy.key]),
        : ascending(a[sortBy.key], b[sortBy.key])
    })
  }
```
or is there something wrong with them?

src/components/Chart/Table.js
```diff
@@ -0,0 +1,255 @@
...
+  // helper function that toggles order (desc/asc) or sets new sort by key (order: desc)
+  const setSort = key => {
+    if (sortBy.key === key) {
+      setSortBy({
+        key,
+        order: sortBy.order === 'desc' ? 'asc' : 'desc',
+        enableSorting: true
+      })
+    } else {
+      setSortBy({ key, order: 'desc', enableSorting: true })
```
tpreusse, line 108:
```suggestion
      setSortBy({ key, order: 'desc' })
```

src/components/Chart/Table.js
```diff
@@ -0,0 +1,255 @@
...
+  })
+
+  // helper function that toggles order (desc/asc) or sets new sort by key (order: desc)
+  const setSort = key => {
+    if (sortBy.key === key) {
+      setSortBy({
+        key,
+        order: sortBy.order === 'desc' ? 'asc' : 'desc',
+        enableSorting: true
+      })
```
tpreusse, line 106: I think we should disable the sorting again on the third click:
```suggestion
      setSortBy(sortBy.order === 'asc' ? {} : {
        key,
        order: sortBy.order === 'desc' ? 'asc' : 'desc'
      })
```
(disabled === empty object === undefined sortBy.key)

src/components/Chart/Table.js
```diff
@@ -0,0 +1,255 @@
...
+      }
+    } else {
+      scale = scaleOrdinal()
+      domain = parsedData.map(d => d[column])
+      currentColorRange = colorRanges.discrete.slice(0, domain.length + 1)
+    }
+    return scale
+      .domain(domain)
+      .range(currentColorRange || colorRanges.sequential)
+  }
```
tpreusse, line 136: this is a bit crazy: we create a scale per column? this could mean we'd have two different color scales in action in the same table? one for number columns and an ordinal for string columns? And only the `number` columns support a colorRange defined outside? And no support for `colorMap` at all?

IMO just create one color scale. If `thresholds` is set use `scaleThreshold`, if not `scaleOrdinal`. If you want `scaleQuantize` support first add an example and then maybe set it behind the following condition: `tableColumns.filter(c => c.color).every(c => c.type === 'number')`

src/components/Chart/Table.js
```diff
@@ -0,0 +1,255 @@
...
+        <thead>
+          <tr>
+            {columns.map((tableHead, index) => (
+              <th
+                {...styles.header}
+                {...colorScheme.set('borderBottomColor', 'text')}
+                style={{
+                  textAlign:
+                    tableColumns.find(d => d.column === tableHead)?.type ===
+                    'number'
```
tpreusse, line 150:
```suggestion
                   numberColumns.includes(tableHead)
```

src/components/Chart/Table.js
```diff
@@ -0,0 +1,255 @@
...
+          </tr>
+        </thead>
+        <tbody style={{ marginTop: '5px' }}>
+          {sortedData.map((row, rowIndex) => (
+            <tr
+              key={rowIndex}
+              {...(rowIndex % 2 !== 0 &&
+                colorScheme.set('backgroundColor', 'hover'))}
+            >
+              {Object.keys(row).map((cellKey, cellIndex) => (
```
tpreusse, line 177:
```suggestion
          {parsedData.map((row, rowIndex) => (
            <tr
              key={rowIndex}
              {...(rowIndex % 2 !== 0 &&
                colorScheme.set('backgroundColor', 'hover'))}
            >
              {columns.map((cellKey, cellIndex) => (
```

src/components/Chart/Table.js
```diff
@@ -0,0 +1,255 @@
...
+  customThreshold: PropTypes.number
+}
+
+Table.defaultProps = defaultProps.Table
+
+Table.propTypes = propTypes
+
+export default Table
+
+const Cell = props => {
```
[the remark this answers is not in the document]
felixmichel: i do this because I was thinking of Bar and Sparklines as other components later

src/components/Chart/Table.js
```diff
@@ -0,0 +1,255 @@
...
+    colorScale,
+    value,
+    column,
+    children
+  } = props
+  return (
+    <td
+      {...style}
+      style={{
+        width: width + 'px',
```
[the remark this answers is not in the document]
felixmichel: sweet

src/components/Chart/Table.js
felixmichel: what would be the domain then for scaleOrdinal?

tpreusse: Something like this:
```
const colorColumns = tableColumns.filter(c => c.color).map(c => c.column)
const colorValues = !colorColumns.length ? [] : parsedData.reduce(
  (values, row) => {
    colorColumns.forEach(key => {
      values.push(row[key])
    })
    return values
  }
  []
).filter(Boolean).filter(deduplicate)

const color = getColorMapper(props, colorValues)
```

src/components/Chart/Table.js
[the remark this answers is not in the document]
tpreusse: If you want you can leave it but my hunch is that `TableCellBar` and `TableCellSparkline` would be the things that go in separate components.

tpreusse:
`th` seem to be missing horizontal padding:
Screenshot 2022-01-21 at 10 54 47

src/components/Chart/Table.js
```diff
@@ -0,0 +1,228 @@
...
+                  type={tableColumns.find(d => d.column === cellKey)?.type}
+                  width={tableColumns.find(d => d.column === cellKey)?.width}
+                  color={tableColumns.find(d => d.column === cellKey)?.color}
+                  value={row[cellKey]}
+                  colorScale={colorScale}
+                >
+                  {tableColumns.find(d => d.column === cellKey)?.type ===
+                  'number'
+                    ? getFormat(numberFormat)(row[cellKey])
+                    : row[cellKey]}
```
tpreusse, line 182:
```suggestion
                  {numberColumns.includes(tableHead)
                    ? getFormat(numberFormat)(row[cellKey])
                    : row[cellKey]}
```

and move the `getFormat(numberFormat)` outside of render of individual cells as it only needs to be done once

src/components/Chart/Table.js
```diff
@@ -0,0 +1,228 @@
...
+          ))}
+        </tbody>
+      </table>
+    </div>
+  )
+}
+
+export const propTypes = {
+  values: PropTypes.array.isRequired,
+  tableColumns: PropTypes.array,
```
tpreusse, line 195: would be nice to type this out with `PropTypes.shape`

src/components/Chart/ChartContext.js
```diff
@@ -277,6 +277,13 @@ export const defaultProps = {
     inlineLabelThreshold: 10,
     padding: 0,
     colorMap: 'swissPartyColors'
+  },
+  Table: {
+    color: 'label',
+    colorRange: 'sequential',
```
tpreusse, line 283:
```suggestion
```

src/components/Chart/Table.docs.md
```diff
@@ -0,0 +1,148 @@
...
+```react
+<div>
+  <ChartTitle>Öffentlich, privat – oder beides?</ChartTitle>
+  <ChartLead>Welche Angebote in den einzelnen europäischen Ländern dominieren</ChartLead>
+  <CsvChart
+    config={{
+      "type": "Table",
+      "colorRange": [
+        "#762a83", "#af8dc3", "#e7d4e8", "#d9f0d3", "#7fbf7b", "#1b7837"
+        ],
```
tpreusse, line 99:
```suggestion
```

src/components/Chart/Table.js
```diff
@@ -0,0 +1,228 @@
...
+
+  if (thresholds) {
+    domain = thresholds
+    if (!colorRange) {
+      currentColorRange = colorRanges.sequential.slice(0, domain.length + 1)
+    }
+    colorScale = scaleThreshold()
+      .domain(domain)
+      .range(currentColorRange || colorRanges.sequential)
+  } else {
```
tpreusse, line 115:
```suggestion
  let colorScale

  if (thresholds) {
    colorScale = scaleThreshold()
      .domain(thresholds)
      .range(
        colorRanges[colorRange] ||
        colorRange ||
        colorRanges.sequential.slice(0, thresholds.length + 1)
      )
  } else {
```

tpreusse (commented):
thx & awesome work

one last round of making the code shorter and nicer ;)

src/components/Chart/Table.js
```diff
@@ -0,0 +1,233 @@
...
+  const {
+    values,
+    numberFormat,
+    colorRanges,
+    colorRange,
+    defaultSortColumn,
+    thresholds,
+    tableColumns
+  } = props
+  const columns = values.columns
```
tpreusse, line 57: One more thing:
```suggestion
  const columns = values.columns || Object.keys(values[0] || {})
```

This is for the case that values was not parsed with `d3-dsv` which provides `.columns` or is empty array like the default prop.

src/components/Chart/Table.js
```diff
@@ -0,0 +1,233 @@
...
+export default Table
+
+const Cell = props => {
+  const { type, width, color, colorScale, value, children } = props
+  return (
+    <td
+      {...(type === 'number' && styles.cellNumber)}
+      {...styles.cell}
+      style={{
+        width: +width,
```
tpreusse, line 225: OOOps two more things:

Screenshot 2022-01-21 at 16 24 40


```suggestion
        width: width !== undefined ? +width : undefined,
```

